PR #49, opened 2023-08-10
Create stale.yml

daogrady:
See https://github.com/actions/stale#all-options for more options.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,32 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: '36 18 * * *'
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 14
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 14
```
bugwelle, line 24: 2 weeks is not a long time. As far as I can see, you are the main contributor. A two week vacation isn't unlikely. :smile:

bugwelle (approved):
Thanks for considering my input. As mentioned, I'm not a big fan of the stale bot and would prefer the no-response bot. However, I get why you chose the stale-bot.

Options I'd love to see:

- [exempt-all-milestones](https://github.com/actions/stale#exempt-all-milestones)
  If you work with milestones, I wouldn't close issues that you put on a milestone. If they are closed, it could give the false sensation that the issue/feature request has been fixed/implemented.
- [exempt-issue-labels](https://github.com/actions/stale#exempt-issue-labels) (+PR)
  Do you have and keep feature requests in the issue tracker?  If so, I don't think they should be automatically closed.

Maybe if I ever get the time to do so, I will try to update the no-response bot and add a version on the cap-js org or something, that can be used by others.

I still don't like the stale bot, because it can easily happen that actual bug reports are closed, because the issue author has given up. On the other hand, this repository doesn't have many issues and I think you are not just fast but also thorough with replying to issues here. I doubt that what I've seen in other repositories will happen to yours. :-)

daogrady:
Thank you for your feedback (and the praise 😉)! Those are all very valid points and I have updated the workflow accordingly.
If you ever get around to updating no-response and feel like it brings added value to this repo, then do let me know -- I am totally open to switching over when expedient.

PR #67, opened 2023-09-05
Upgrading typedoc

danjoa:
Using `^` with `0.` releases never upgrades to new ones → always prefer using `>=` instead.
And because of that, we were not able to upgrade `typescript` to the latest version `5.2.2`. 

Better even: I'd prefer to not have `typedoc` being used at all. → We have to manage what 3rd party/OSS tools we use, otherwise `cap/dev` exceeds all limits, getting ever more heavyweight.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@cap-js/cds-typer",
-  "version": "0.7.0",
+  "version": "0.8.0",
   "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


chgeo (approved):

daogrady:
> Better even

Then I have great news! As the `typedoc` dep was an artefact of providing generated HTML docs, which we decided to not use in the end, I have removed the dependency altogether.

Issue #121, opened 2023-12-06
[QUESTION] Aren't types generated for a view?


mcastrup:
### Question

I noticed that for some of our entities no types are generated. It seems that those entities correspond to views, not to tables.

Example:
The file [Cryptocurrencies.cds](https://github.tools.sap/erp4sme/crypto-for-business/blob/update_cds2types/db/core/cryptocurrencies/Cryptocurrencies.cds) contains the entities

- `CryptocurrenciesBuiltIn`
- `CryptocurrenciesBuiltInConfig`
- `CryptocurrenciesCustom`
- `Cryptocurrencies`

The first three entities are "normal" HANA tables. The fourth one is a union of the first three entities, like:
```
entity Cryptocurrencies                 as
        select from CryptocurrenciesCustom {
            key ID,
                code,
                name,
                ...
        }
    union all
        ( select from CryptocurrenciesBuiltIn as builtIn
            left outer join CryptocurrenciesBuiltInConfig as conf
                on conf.Cryptocurrency.ID = builtIn.ID
            {
                builtIn.ID,
                builtIn.code,
                builtIn.name,
                ...
            }
        );
```

As you can see in the generated files [index.js](https://github.tools.sap/erp4sme/crypto-for-business/blob/update_cds2types/%40cds-models/sap/erp4sme/c4b/core/cryptocurrencies/index.js) and [index.ts](https://github.tools.sap/erp4sme/crypto-for-business/blob/update_cds2types/%40cds-models/sap/erp4sme/c4b/core/cryptocurrencies/index.ts), only types for the first three entities are created. However, we also refer in our coding to the `Cryptocurrencies` entity and need also types for it. Why are these types missing?

Best regards,
Michael

daogrady:
Hi Michael,

thank you for pointing out this use case! Entities that are based on queries indeed don't receive any treatment from cds-typer so far.
I will look into the matter.

Best,
Daniel

mcastrup:
@daogrady 
Thanks for the fast reply. Indeed, it would be great if cds-typer can also cover views. cds2types, which we currently use, generates interfaces also for views.
I also noticed that VSC shows some errors in the generated index.ts files, see the screen shot. I think this is also related to the missing artefatcs.
![image](https://github.com/cap-js/cds-typer/assets/153120429/77c2c427-37f6-4af5-98e7-60da62961f85)

daogrady:
Hi @mcastrup ,

[...]

PR #123, opened 2023-12-12
Generate View Types

daogrady:
Fixes https://github.com/cap-js/cds-typer/issues/121

```cds
entity Foo {
  a: Integer;
  b: String;
}

entity Bar as select from Foo {
  a,
  a as alias
};
```
⬇️ (more or less)
```ts
class Foo {
  a: number
  b: string
}

class Bar {
  a: number
  alias: number
}
```


mcastrup:
Seems to work 👍

Issue #137, opened 2024-01-02
[BUG] enum value resolution fails for actions


fabiantschirschnitz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

Hey @daogrady,

I hope you had a great christmas and a good start into 2024!

It seems that https://github.com/cap-js/cds-typer/commit/133273d0f6f4e83696ebbd64ff754556f37ad7c5
introduced a bug on our side starting with 0.15.0 of cds-typer. The generated typescript code is malicious, since it emits the string enum values as types that are not generated by cds-typer.

### Expected Behavior

String values are  emitted or just the enum type name as in 0.14.0

### Steps To Reproduce

This is our enum type:
```cds
type LockActivityType           : String(20) enum {
    disconnection          = 'DISCONNECTION';
    // TODO add more types
    reconnection           = 'RECONNECTION'
}
```
And here we use it in an action as in-parameter:
```cds
action CreateActivity(in : $self,
                            @mandatory
                            type : db.LockActivityType,
```

The emitted Typescript looks following:
```typescript
CreateActivity: { (type: DISCONNECTION | RECONNECTION, planedDate: string | null, autoTourAssignment: boolean | null, tourID: string | null): any, __parameters: {type: DISCONNECTION | RECONNECTION, planedDate: string | null, autoTourAssignment: boolean | null, tourID: string | null}, __returns: any }
```

This leads to tsc-errors, since `DISCONNECTION` and `RECONNECTION` are no types.

### Environment

```markdown
- **OS**: win11
- @sap/cds: 7.5.1
- Node.js: v18.16.1
- **cds-typer**: 0.15.0
```

### Repository Containing a Minimal Reproducible Example

sorry, not this time. but please let me know, if the error is till unclear :)

### Anything else?

fyi @nlaenger

daogrady:
Hi Fabian,

thanks for reporting this issue! It does seem like the quotes around string enums have gone missing for this constellation. I will look into it.

Best,
Daniel

PR #141, opened 2024-01-10
Resolve externally defined enums in action parameters

ogrady:
Fixes https://github.com/cap-js/cds-typer/issues/137

Using an externally (read: not inline) defined enum as type for an action parameter would incorrectly assume non-string values. This is now fixed by instead resolving the enum's name.

### DIFF
--- a/lib/components/enum.js
+++ b/lib/components/enum.js
@@ -59,7 +59,7 @@ const stringifyEnumType = kvs => [...uniqueValues(kvs)].join(' | ')
 const uniqueValues = kvs => new Set(kvs.map(([,v]) => v?.val ?? v))  // in case of wrapped vals we need to unwrap here for the type
 
 // in case of strings, wrap in quotes and fallback to key to make sure values are attached for every key
-const enumVal = (key, value, enumType) => enumType === 'cds.String' ? JSON.stringify(`${value ?? key}`) : value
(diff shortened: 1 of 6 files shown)


lib/components/enum.js
```diff
@@ -59,7 +59,7 @@ const stringifyEnumType = kvs => [...uniqueValues(kvs)].join(' | ')
 const uniqueValues = kvs => new Set(kvs.map(([,v]) => v?.val ?? v))  // in case of wrapped vals we need to unwrap here for the type
 
 // in case of strings, wrap in quotes and fallback to key to make sure values are attached for every key
-const enumVal = (key, value, enumType) => enumType === 'cds.String' ? JSON.stringify(`${value ?? key}`) : value
+const  enumVal = (key, value, enumType) => enumType === 'cds.String' ? JSON.stringify(`${value ?? key}`) : value
```
fabiantschirschnitz, line 62: :)

fabiantschirschnitz (commented):
Thank you! @daogrady

lib/components/enum.js
daogrady: not sure I understand 🤔
Is the proposed change a single additional space?

fabiantschirschnitz: No, i was just wondering about the extra added space :)

PR #166, opened 2024-02-26
Projection on view does not have correct fields: Unit tests

nils:
This PR does _not_ contain a solution yet, but contains testcases for issue #167.

### DIFF
--- a/test/unit/files/views/model.cds
+++ b/test/unit/files/views/model.cds
@@ -12,3 +12,9 @@ entity FooView as
         code,
         code as alias
     };
+
+entity FooViewProjection as
+    projection on FooView {
+        id as IdProjected,
+        alias as AliasProjected
+    };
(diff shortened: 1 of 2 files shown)


daogrady:
@nils just to make sure: are you planning on implementing the fix for this on top, or is the scope for this PR just the tests? In the latter case, is it ready for review or still in draft?

daogrady:
@nils no worries, navigating through CSN can be quite the task and wrt projections the solution was not exactly straight-forward either. :)
I decided to [extend](https://github.com/cap-js/cds-typer/pull/175/files#diff-a1a8241661e273eff382389573cf54ca241822b944822e2f697120d04177a07d) a bit on the test suite you offered to cover some additional tests, like renaming and only selecting a subset of properties. If you find that these tests cover your case we can supersede your original PR (regretfully, as the provision of runnable test cases was highly appreciated!).

nils:
Thanks @daogrady for the efforts you put into this! 
I have not tested your changes yet, but looking at the test suite the solution should also cover our issue.

I don't mind my testing code becoming obsolete, it served its purpose nevertheless 🙂 Your extended tests look great to me!

Issue #163, opened 2024-02-21
`Plural.drafts` could be `typeof Plural`


daogrady:
Currently, both the singular and the plural form of entities have a reference to `.drafts`. This makes sense as they point to the same CSN entity during runtime and should be treated the same. So:

```ts
class Singular {
  static drafts: typeof Singular
}

class Plural extends Array<Singular> {
  static drafts: typeof Singular
}
```

This leads to confusion when the type should not actually matter:

```ts
SELECT.from(b ? Plural : Plural.drafts)  // Singular | Plural -> no proper type support
```

Find out if there is a compelling argument against

```ts
class Plural extends Array<Singular> {
  static drafts: typeof Plural
}
```

and if not implement this behaviour.

geert-janklaps:
Hi @daogrady,

This actually is causing issues when registering handlers, if you register the handler with the plural of the drafts it's typed as an object. At runtime the value is not an object but an array.
![image](https://github.com/cap-js/cds-typer/assets/2744548/24d3ee2e-a3bd-47d9-ad89-f8077318f0f3)

Cheers,
Geert-Jan

geert-janklaps:
Pull request #192 created to resolve this issue. (ran unit tests & did the same manual change in my running project to validate)

PR #192, opened 2024-04-04
fix: plural drafts #163

geert-janklaps:


daogrady:
Hi @geert-janklaps,

thank you very much for taking on this issue! I have added a changelog entry and test cases for your contribution -- especially the latter are a bit prickly to write.
I will discuss this change with the runtime team later this day and if they don't object we can include this in the next release.
(FYI @johannes-vogel )

Best,
Daniel

daogrady:
Update on that: we still need to streamline how and where we want to allow the singular/ plural semantic, so I would like to postpone this PR until we have clarity on that.

geert-janklaps:
Hi @daogrady,

Sounds good to me! Do you have an idea when this could become part of the release / how long you'd postpone?
Currently I need to perform this change in the node modules folder manually for the handler registration to work (and not fail due to typing issues). This also means that for one of our solutions I'm unable to setup CI/CD pipelines for automated build and deployments. (nothing we can't handle, but it would make life easier for sure)

Cheers,
Geert-Jan

geert-janklaps:
Hi @daogrady,

Any update on this one by any chance? (mainly commenting to prevent this from being closed automatically)

Cheers,
Geert-Jan

daogrady:
Hi Geert-Jan,

sorry for the long silence on this one! The blocking issues have been resolved afaik, so I will bring this up again in the next cds-types meeting.

Best,
Daniel

P.S.: bummer I didn't run into you at recap 😕

geert-janklaps:
> Hi Geert-Jan,
> 
> I have not forgotten about this, but finding the right solution here is non-trivial. On the one hand, it would be correct to use a plural type for the `.drafts` on plural. On the other hand, it would mess with how types are derived in handler registrations, so you will probably get `Array<Array<Singular>>` in some cases. This needs some more thinking through, I believe. Sorry for the wait. :(
> 
> Best, Daniel

Hi @daogrady,

No worries, in this case there are workarounds as suggested by Johannes above. I might be overlooking some cases (I'm looking at this from an implementation point of view), so definitely take your time to go through details so this doesn't create new bugs :)

Cheers,
Geert-Jan

daogrady:
keep alive

stockbal:
Hi @daogrady,

I enhanced the tests for CQL and service handler registration for plural drafts. See [cds-services.ts](https://github.com/stockbal/cds-types/blob/01d8b7730dabcaf324360a6335efb08b4d2bf4d5/test/typescript/apis/project/cds-services.ts#L236-L283) and [cds-ql.ts](https://github.com/stockbal/cds-types/blob/01d8b7730dabcaf324360a6335efb08b4d2bf4d5/test/typescript/apis/project/cds-ql.ts#L93-L123).

I could find only one issue when registering the handler for `on("READ")`, but it was actually not because of the plural draft, but the singular one 😅, which infers that the `data` parameter in the handler is a singular object. I explained this issue in the test case for `cds-services.ts`.

Please have a look when you have some time. From what I can see, nothing would stand in the way of making the switch of Plural.drafts from `typeof Singular` to `typeof Plural`.

Regards,
Ludwig

daogrady:
Hi Geert-Jan,

sorry for the very belated feedback. `.drafts` was just something that was continuously displaced by other tasks, so I was not able to give you an appropriate review in due time.
I just noticed that the underlying [issue](https://github.com/cap-js/cds-typer/issues/163) was actually marked as fixed in another PR. I hope your concerns are thereby addressed and cds-typer now produces `.draft` properties as you'd expect them to behave.

Best,
Daniel

Issue #164, opened 2024-02-21
Add `CdsDate` and `CdsDateTime` types


daogrady:
The types `Date` and `DateTime` from CDS are represented as _yyyy-mm-dd-..._ strings during runtime. So they currently get typed to `string`. 
This has led to [confusion](https://github.com/cap-js/cds-typer/issues/157) in the past.
While there is now a section addressing this in [the FAQ](https://github.com/cap-js/cds-typer/wiki/FAQ), it would be great to understand this choice from looking at the generated types.

We should therefore have two types

```ts
type CdsDate = string
type CdsDateType = string
```

with appropriate explanation in their doc. These definitions can be part of the [base definitions](https://github.com/cap-js/cds-typer/blob/main/lib/file.js#L559-L586) that are included in every typer run. The CDS types `Date` and `DateTime` can then [map to](https://github.com/cap-js/cds-typer/blob/main/lib/components/resolver.js#L59-L60) these new types instead.

PR #193, opened 2024-04-05
Add date types

alyonavyshnevska:
Added four new types:
- `CdsDate`
- `CdsTme`
- `CdsDateTime`
- `CdsTimeStamp`

which correspond to the built-in types `Date`, `Time`, `DateTime` and `Timestamp` in `cds`. 

During runtime, these types are just strings. `CdsDate` and `CdsTime` each come with a regex that only accepts values in the same format as `cds.Date` and `cds.Time`. 
 
resolves https://github.com/cap-js/cds-typer/issues/164

- [x] changelog


lib/file.js
```diff
@@ -634,6 +637,15 @@ export type EntitySet<T> = T[] & {
...
 } & Exclude<Required<T>, null>;
+
+
+/**
+ * Dates and timestamps are strings during runtime, so cds-typer represents them as such.
+ */
+export type CdsDate = ${dateRegex};
+export type CdsDateTime = string;
+export type CdsTime = ${timeRegex};
+export type CdsTimestamp = string;
```
daogrady, line 648:
```suggestion
export type CdsTimeStamp = string;
```

daogrady (changes requested):
Looks mostly good, just the spelling on the generated type is slightly out of line with how it is generated in the builtins (which leads to an "unresolved type" error in the generated files). See my comments.

lib/file.js
alyonavyshnevska: ah, actually this should be the other way around to be consistent with the [built-in types](https://pages.github.tools.sap/cap/docs/guides/domain-modeling#standard-built-in-types). Good that you noticed the discrepancy there.

lib/file.js
```diff
@@ -634,6 +637,15 @@ export type EntitySet<T> = T[] & {
 export type DeepRequired<T> = { 
     [K in keyof T]: DeepRequired<T[K]>
 } & Exclude<Required<T>, null>;
+
+
+/**
+ * Dates and timestamps are strings during runtime, so cds-typer represents them as such.
+ */
+export type CdsDate = ${dateRegex};
```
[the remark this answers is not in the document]
alyonavyshnevska: My mistake, did not commit the change

daogrady (approved):
Looks good now, thanks for the contribution! :)

PR #233, opened 2024-05-08
Add build plugin

daogrady:
Adds a plugin for `cds build` to build TypeScript projects. Can be called explicitly using `cds build --for typescript`.
This plugin features two modes:

### Automatic
Assumes a standard project setup. Will trigger transpilation process and copy files to the correct positions. Will also try to copy types for models to the _gen/_ folder for consistency.

### Configured
Users can create a _tsconfig.cdsbuild.json_ which is passed to the transpilation step.


cds-plugin.js
```diff
@@ -0,0 +1,125 @@
...
+        return '@cds-models'
+    }
+
+    init() {
+        this.task.dest = path.join(cds.root, 'gen', 'srv')
+    }
+
+    async #runCdsTyper () {
+        DEBUG?.('running cds-typer')
+        await exec(`npx @cap-js/cds-typer "*" --outputDirectory ${this.#modelDirectoryName}`)
```
LotharBender, line 80: Wouldn't it make sense to have a cds-typer API that could be called from both the build plugin and the typer CLI?

`this.model()` returns a compiled CSN representing the correct transitive closure of the model. 
I assume that the typer generates types for all entities and type definitions. 
The corresponding model files and folders are either determined by the cds build using `cds.resolve('*')` or defined by the user as part of the `build task configuration`. 
On the one hand, this would ensure that the same default model scope is used as for other build tasks. On the other hand, users might expect the typer to use the given model options as configured in the build task.

LotharBender (changes requested):

LotharBender (approved):
Sorry, wanted to approve...

cds-plugin.js
daogrady: Very good point! While I would prefer to just feed `this.model()` into the typer api, this is currently not possible, as cds-typer requires two flavours of CSN (xtended and inferred). So we'll still need to read the model from the files, but we at least got rid of the spawned subprocess. 👍

PR #234, opened 2024-05-08
Streamlined eslint usage

danjoa:


daogrady:
As discussed, cds-typer and cds-types require their own lint config as they are maintained separately from our other tooling. I will review the current config to see if some of the current rules are covered by defaults at a later point in time.

PR #241, opened 2024-05-17
adjust for 'Name.Name' entity names (namespace equals entity name)

nils:
Slight adjustment of the 'scoped' unit test to also cover situations where the namespace/service name equals the prefix of a contained entity. Before this, the entity name was shortened/stripped of its prefix/name.


daogrady:
Hi Nils,

good catch, thanks for the contribution! Could you please also add a matching entry in the [CHANGELOG.md](https://github.com/cap-js/cds-typer/blob/main/CHANGELOG.md)?

Best,
Daniel

nils:
Hi @daogrady,

thanks for the review! I have added the changelog entry accordingly.

daogrady:
Appreciated, thanks again! 🚀

daogrady (approved):

Issue #253, opened 2024-06-04
[BUG] cds build plugin throws error


Kyrodan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

JavaScript, TypeScript

### Current Behavior

Since v0.21.1 and #248 the plugion is now bundled.
Running `cds build`now throws an exception:
`Error: Illegal priority for  encountered for custom plugin - in this version only priority value '1' is allowed`

### Expected Behavior

Doen't throws exception.

### Steps To Reproduce

run `cds build`

### Environment

```markdown
- **OS**: Windows 10
- **Node**: v20.13.1
- **npm**: 10.8.0
- **cds-typer**: 0.21.1
- **cds**: cds-dk [7.9.2], cds [7.9.2], compiler [4.9.4]
```

### Anything else?

From @sap\cds-dk\lib\build\buildTaskEngine.js:214:19:
![grafik](https://github.com/cap-js/cds-typer/assets/355178/5ed2810c-017d-4c54-beca-90f23c7f863c)

daogrady:
Hi Daniel,

thanks for pointing this out! This happens because I had expected cds8 to be available by now (which is required for this build task to work).
So this is currently a rather unfortunate error message that will automagically fix itself as soon as you are using cds8 (yet to be released!).
Sorry for the inconvenience!

Best,
Daniel

daogrady:
Starting with the upcoming release of cds-typer, the build task will only attempt to run if cds8 is in place to not confront users with this ugly error.
Again, once cds8 is released and in use this problem will go away, but I see how the current output is confusing to users. Thanks again for pointing it out!

PR #254, opened 2024-06-05
Disable typescript buildtask in cds7

daogrady:
As cds8 is not released as of today, users are [encountering](https://github.com/cap-js/cds-typer/issues/253) undesired behaviour, as cds7 does not allow for custom plugins to run after builtin ones.
This PR effectively disables the typescript build task altogether for cds <8.
It is silently removed from `cds build` and raises an error if the user tries to force it via `cds build --for typescript`.


cds-plugin.js
```diff
@@ -47,13 +47,16 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
     )
     : undefined
 
+// FIXME: remove once cds7 has been phased out
+const isAtLeastCds8 = cds?.version && cds.version >= '8.0.0'
```
LotharBender, line 51: Why not using:
`if ( cds?.version && cds.version < '8.0.0') return`

Note: in general using `>` `<` operators for semver versions comparing might produce wrong results, but here it should be ok ;-)

daogrady: isn't that just the De Morgan'd version of what I used? Or is there a difference/ advantage I am overlooking? :D

LotharBender: - Yes, I used the negotiation to return immediately, you can of course also use your version :) 
My point was - why not check once and return immediately without registering the build plugin which won't work anyway?

LotharBender (approved):

LotharBender (approved):

PR #281, opened 2024-07-23
feat: use `Proxy`s for entity exports in index.js files

stockbal:
Wrapping the `module.exports` in `Proxy` objects removes the need for dynamic imports everywhere.

### Changes

- new `module.exports` in `index.js` of namespace `_`  
   image
- wrapping the entity `module.exports` in the new `createEntityProxy` function  
  image

### Remarks

- The `set`-trap in the Proxy and the `customProps` are mainly for the inline enums, otherwise they could not be accessed

### DIFF
--- a/lib/components/javascript.js
+++ b/lib/components/javascript.js
@@ -3,18 +3,18 @@
 const proxyAccessFunction = function (namespace, name, opts = {}) {
     const { target, customProps } = { target: {}, customProps: [], ...opts }
     const fq = namespace !== '' ? `${namespace}.${name}` : name
-    let _entity
+    let entity
     return new Proxy(target, {
         get: function (target, prop) {
             // we already know the name so we skip the cds.entities proxy access
             if (prop === 'name') return fq
             // custom properties access on 'target' as well as cached _entity property access goes here
-            if (Object.prototype.hasOwnProperty.call(target, prop)) return target[prop]
+            if (Object.hasOwn(target, prop)) return target[prop]
             // inline enums have to be caught here for first time access, as they do not exist on the entity
-            if (customProps.length && customProps.includes(prop)) return target[prop]
+            if (customProps.includes(prop)) return target[prop]
             // last but not least we pass the property access to cds.entities
             if (cds.entities) {
-                const entity = _entity ? _entity : (_entity = cds.entities(namespace)[name])
+                entity ??= cds.entities(namespace)[name]
                 if (!entity) throw new Error(`'${fq}' could not be found in cds.entities('${namespace}')`)
                 if (entity[prop]) target[prop] = entity[prop]
                 return target[prop]
--- a/lib/components/javascript.js
+++ b/lib/components/javascript.js
@@ -19,7 +19,7 @@ const proxyAccessFunction = function (fqParts, opts = {}) {
             // inline enums have to be caught here for first time access, as they do not exist on the entity
             if (customProps.includes(prop)) return target[prop]
             // last but not least we pass the property access to cds.entities
-            throw new Error(`Property ${prop} does not exist on entity '${fq}' or cds.entities is not yet defined`)
+            throw new Error(`Property ${prop} does not exist on entity '${fq}' or cds.entities is not yet defined. Ensure the CDS runtime is fully booted before accessing properties.`)
         }
     })
 }.toString()
--- a/lib/components/javascript.js
+++ b/lib/components/javascript.js
@@ -1,13 +1,14 @@
 /* eslint-disable no-undef */
 // this function will be stringified as part of the generated types and is not supposed to be used internally
-const proxyAccessFunction = function (namespace, name, opts = {}) {
+// @ts-nocheck
+const proxyAccessFunction = function (fqParts, opts = {}) {
     const { target, customProps } = { target: {}, customProps: [], ...opts }
-    const fq = namespace !== '' ? `${namespace}.${name}` : name
-    let entity
+    const fq = fqParts.join('.')
     return new Proxy(target, {
         get: function (target, prop) {
             if (cds.entities) {
                 target.__proto__ = cds.entities[fq]
+                // overwrite/simplify getter after cds.entities is accessible
                 this.get = (target, prop) => target[prop]
                 return target[prop]
             }
--- a/lib/components/javascript.js
+++ b/lib/components/javascript.js
@@ -20,10 +20,6 @@ const proxyAccessFunction = function (namespace, name, opts = {}) {
                 return target[prop]
             }
             throw new Error(`Property ${prop} does not exist on entity '${fq}' or cds.entities is not yet defined`)
-        },
-        set: function (target, prop, newVal) {
-            target[prop] = newVal
-            return true
         }
     })
 }.toString()
--- a/lib/components/javascript.js
+++ b/lib/components/javascript.js
@@ -14,7 +14,7 @@ const proxyAccessFunction = function (namespace, name, opts = {}) {
             if (customProps.includes(prop)) return target[prop]
             // last but not least we pass the property access to cds.entities
             if (cds.entities) {
-                entity ??= cds.entities(namespace)[name]
+                entity ??= cds.entities[fq]
                 if (!entity) throw new Error(`'${fq}' could not be found in cds.entities('${namespace}')`)
                 if (entity[prop]) target[prop] = entity[prop]
                 return target[prop]
--- a/lib/components/javascript.js
+++ b/lib/components/javascript.js
@@ -1,0 +1,32 @@
+/* eslint-disable no-undef */
+// this function will be stringified as part of the generated types and is not supposed to be used internally
+const proxyAccessFunction = function (namespace, name, opts = {}) {
+    const { target, customProps } = { target: {}, customProps: [], ...opts }
+    const fq = namespace !== '' ? `${namespace}.${name}` : name
+    let _entity
+    return new Proxy(target, {
+        get: function (target, prop) {
+            // we already know the name so we skip the cds.entities proxy access
+            if (prop === 'name') return fq
+            // custom properties access on 'target' as well as cached _entity property access goes here
+            if (Object.prototype.hasOwnProperty.call(target, prop)) return target[prop]
+            // inline enums have to be caught here for first time access, as they do not exist on the entity
+            if (customProps.length && customProps.includes(prop)) return target[prop]
+            // last but not least we pass the property access to cds.entities
+            if (cds.entities) {
+                const entity = _entity ? _entity : (_entity = cds.entities(namespace)[name])
+                if (!entity) throw new Error(`'${fq}' could not be found in cds.entities('${namespace}')`)
+                if (entity[prop]) target[prop] = entity[prop]
+                return target[prop]
+            }
+            throw new Error(`Property ${prop} does not exist on entity '${fq}' or cds.entities is not yet defined`)
+        },
+        set: function (target, prop, newVal) {
+            target[prop] = newVal
+            return true
+        }
+    })
+}.toString()
+/* eslint-enable no-undef */
+
+module.exports = { proxyAccessFunction }
--- a/lib/components/javascript.js
+++ b/lib/components/javascript.js
@@ -6,6 +6,11 @@ const proxyAccessFunction = function (namespace, name, opts = {}) {
     let entity
     return new Proxy(target, {
         get: function (target, prop) {
+            if (cds.entities) {
+                target.__proto__ = cds.entities[fq]
+                this.get = (target, prop) => target[prop]
+                return target[prop]
+            }
             // we already know the name so we skip the cds.entities proxy access
             if (prop === 'name') return fq
             // custom properties access on 'target' as well as cached _entity property access goes here
@@ -13,12 +18,6 @@ const proxyAccessFunction = function (namespace, name, opts = {}) {
             // inline enums have to be caught here for first time access, as they do not exist on the entity
             if (customProps.includes(prop)) return target[prop]
             // last but not least we pass the property access to cds.entities
-            if (cds.entities) {
-                entity ??= cds.entities[fq]
-                if (!entity) throw new Error(`'${fq}' could not be found in cds.entities('${namespace}')`)
-                if (entity[prop]) target[prop] = entity[prop]
-                return target[prop]
-            }
             throw new Error(`Property ${prop} does not exist on entity '${fq}' or cds.entities is not yet defined`)
         }
     })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## Version 0.26.0 - TBD
 ### Added
+- Added a CLI option `--useEntitiesProxy`. When set to `true`, all entities are wrapped into `Proxy` objects during runtime, allowing top level imports of entity types.
 - Added a static `.kind` property for entities and types, which contains `'entity'` or `'type'` respectively
 - Apps need to provide `@sap/cds` version `8.2` or higher.
 - Apps need to provide `@cap-js/cds-types` version `0.6.4` or higher.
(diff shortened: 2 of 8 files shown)


daogrady:
Hi Ludwig,

thank you for this contribution. I gave it a quick look and it looks promising and well thought-out in my opinion! I like how you considered several edge-cases 👍
I'd would like to set it back a little longer for now, as we wanted to investigate a bit more whether we can accommodate top-level imports on sap/cds level instead of winding our way around in the generated types.

(@chgeo fyi)

On a side note: thanks for providing the (in this case very beneficial) screenshots of the code. It must have been a pain to develop this big chunk of code and put it in and out of the quotes. 🙃 I guess we could in this case do a little no-no and stringify the function to improve qol when working on the proxy itself:

```js
// lib/javascript.js or something like that

// this function will be stringified as part of the generated types and is not supposed to be used internally
const proxyAccessorFunction = function(namespace, name, opts = {}) {
    const { target, customProps } = { target: {}, customProps: [], ...opts }
    const fq = namespace !== '' ? `${namespace}.${name}` : name
    let _entity
    return new Proxy(target, {
        get: function (target, prop) {
            // we already know the name so we skip the cds.entities proxy access
            if (prop === "name") return fq
            // custom properties access on 'target' as well as cached _entity property access goes here
            if (target.hasOwnProperty(prop)) return target[prop]
            // inline enums have to be caught here for first time access, as they do not exist on the entity
            if (customProps.length && customProps.includes(prop)) return target[prop]
            // last but not least we pass the property access to cds.entities
            if (cds.entities) {
                const entity = _entity ? _entity : (_entity = cds.entities(namespace)[name]);
                if (!entity) throw new Error(`'${fq}' could not be found in cds.entities('${namespace}')`)
                if (entity[prop]) target[prop] = entity[prop]
                return target[prop]
            }
            throw new Error(`Property ${prop} does not exist on entity '${fq}' or cds.entities is not yet defined`);
        },
        set: function (target, prop, newVal) {
            target[prop] = newVal
            return true
        }
    })
}.toString()

module.exports = {
  proxyAccessorFunction
}
```

```js
// lib/file
const { proxyAccessorFunction } = require('./javascript.js')

#getEntityProxyFunctionExport() {
  return `module.exports.createEntityProxy = ${proxyAccessorFunction}`
}
```


Best,
Daniel

stockbal:
Hi Daniel,

I like the `toString()` approach for the function. Makes maintenance far easier 😅. Regarding the location of the function, I saw that there already something for TypeScript under `/lib/components`
image

Wouldn't the proxy accessor function qualify as something similar? So we could move it instead to `/lib/components/javasript.js`?

BR,
Ludwig

daogrady:
> Wouldn't the proxy accessor function qualify as something similar? So we could move it instead to /lib/components/javasript.js?

Sure! 🙂

daogrady:
Hi Ludwig,

we wanted to pick that string of work back up, but to our surprise noticed that top level imports did not seem to cause any race condition issues anymore under cds8. We currently assume there has been some change by the runtime colleagues that flew under our radar and fixed the problem on the fly. Can you confirm that top level imports are usable under cds8 with how cds-typer currently generates the files?

Best,
Daniel

stockbal:
Hi Daniel,

I will make some tests and come back to you.

BR,
Ludwig

hakimio:
@daogrady not being able to use cds models in a custom server implementation (without lazy loading them) and jest tests is also an issue in our projects.
I imagine this change would be highly appreciated by many other `cds-typer` users as well.

daogrady:
@stockbal thanks for putting in the effort! I will check what is different from our setup and why we are not seeing the issue on our end.

@hakimio yup 🙂 I bet it's a annoyance to many of our users, as top level imports are something people are used to and expect to be able to use. So this is a change that is very important to us as well.

stockbal:
Hi @daogrady,

any news here?

Regards,
Ludwig

daogrady:
Hi Ludwig,

sorry for the wait, I have not forgotten about your contribution! 😬
We have looked into the matter and not found a better solution so far, so we are very inclined to give your proposal using proxies a shot.
We'd like to investigate two things first, and maybe you have given it some thoughts already:

1. what are the runtime implications? Proxies, in my experience, are rather fast and you have already employed some caching. But in real-world scenarios with large models and frequent access to specific entities and their properties, this could be a hit in time and space complexity. We'd like to make sure the impact stays modest with this change.
2. how conveniently can CSN be debugged when everything is wrapped in `Proxy` objects? Can users still inspect their entities comfortably, or will they have to dive through several layers of JS magic before seeing their actual definitions?

If you have some thoughts on one or both of those issues, feel free to share your findings! If not, I will investigate them asap.

As this is a rather drastic change, we'd like to offer this feature as opt-in if Proxies turn out to be benign solution, to test the waters.

Best,
Daniel

lib/components/javascript.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable no-undef */
+// this function will be stringified as part of the generated types and is not supposed to be used internally
+const proxyAccessFunction = function (namespace, name, opts = {}) {
+    const { target, customProps } = { target: {}, customProps: [], ...opts }
+    const fq = namespace !== '' ? `${namespace}.${name}` : name
+    let _entity
```
daogrady, line 6:
```suggestion
    let entity
```

lib/components/javascript.js
```diff
@@ -0,0 +1,32 @@
...
+        get: function (target, prop) {
+            // we already know the name so we skip the cds.entities proxy access
+            if (prop === 'name') return fq
+            // custom properties access on 'target' as well as cached _entity property access goes here
+            if (Object.prototype.hasOwnProperty.call(target, prop)) return target[prop]
+            // inline enums have to be caught here for first time access, as they do not exist on the entity
+            if (customProps.length && customProps.includes(prop)) return target[prop]
+            // last but not least we pass the property access to cds.entities
+            if (cds.entities) {
+                const entity = _entity ? _entity : (_entity = cds.entities(namespace)[name])
```
daogrady, line 17:
```suggestion
                entity ??= cds.entities(namespace)[name]
```
I guess that should result in the same, no?

lib/components/javascript.js
```diff
@@ -0,0 +1,32 @@
...
+const proxyAccessFunction = function (namespace, name, opts = {}) {
+    const { target, customProps } = { target: {}, customProps: [], ...opts }
+    const fq = namespace !== '' ? `${namespace}.${name}` : name
+    let _entity
+    return new Proxy(target, {
+        get: function (target, prop) {
+            // we already know the name so we skip the cds.entities proxy access
+            if (prop === 'name') return fq
+            // custom properties access on 'target' as well as cached _entity property access goes here
+            if (Object.prototype.hasOwnProperty.call(target, prop)) return target[prop]
```
daogrady, line 12:
```suggestion
            if (Object.hasOwn(target, prop)) return target[prop]
``` 
or is there a specific reason to go via the prototype here?

lib/components/javascript.js
```diff
@@ -0,0 +1,32 @@
...
+            // inline enums have to be caught here for first time access, as they do not exist on the entity
+            if (customProps.length && customProps.includes(prop)) return target[prop]
```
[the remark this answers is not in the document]
daogrady: (https://jsben.ch/xRvAG)

lib/components/javascript.js
stockbal: 1) I did not know about `hasOwn`
2) I wrote the code in typescript and changed it to `prototype` because of the eslint rule https://eslint.org/docs/latest/rules/no-prototype-builtins

lib/components/javascript.js
stockbal: Yes

lib/components/javascript.js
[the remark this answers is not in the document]
stockbal: I guess not unless `customProps` will contain a lot entries, which I don't see happening

stockbal:
Hi Daniel,

here my thoughts about the 2 points you mentioned.

### 1) Performance

I created a small test case to check the impact on multiple calls on the property of one entity. Just a simple case with 1.000.000 successive calls.

#### Accessing `elements` of the proxied entity

```ts
import {MyEntity} from "#cds-models/db";

for (let i = 0; i < 1000000; i++) {
  proxyEntity.elements;
}
```

Duration: ~187ms to ~280ms

#### Accessing `elements` of the entity instance directly

```ts
const entity = cds.entities[MyEntity.name];

for (let i = 0; i < 1000000; i++) {
  entity.elements;
}
```

Duration: ~1ms to ~3ms

### 2) Debugging

Generally speaking one can debug the proxied entities like any other object. The only downside being, that one does not see all properties immediately like before.

image

Until a property like e.g. `elements` is not accessed it is not visible in the variables section. To make it visible one can either create a watch-variable or access it via the debug console. After doing one or the other the `entity`-properties appear under the `[[Target]]` object of the proxy.

### Improvements

Not sure if the added time is negligible or not, but it seemed to high to me, so I experimented a bit and by using the following approach for the proxy function it goes way down to just **~25ms** for 1M calls:

```js
module.exports.createEntityProxy = function (namespace, name, opts = {}) {
  let { target } = { target: {}, ...opts };
  const fq = namespace !== "" ? `${namespace}.${name}` : name;
  let initialized;

  return new Proxy(
    { ...target, name: fq },
    {
      get: function (target, prop) {
        if (!initialized && cds.entities) {
          initialized = true;
          target.__proto__ = cds.entities[fq];
        }
        return target[prop];
      },
      set: function (target, prop, newVal) {
        target[prop] = newVal;
        return true;
      }
    }
  );
};
```
This should also ease the debugging, as all properties will be visible under `[[Target]]` as soon as one property of an entity is accessed. The additional `customProps` is also no longer required.  
The new version behaves more like a delayed version of the classic `module.exports.Books = { __proto__: csn.Books }`.

If you don't see any major downside in the new version, I would suggest we go for that.

Regards,
Ludwig

lib/components/javascript.js
```diff
@@ -0,0 +1,32 @@
...
+                entity ??= cds.entities[fq]
+                if (!entity) throw new Error(`'${fq}' could not be found in cds.entities('${namespace}')`)
+                if (entity[prop]) target[prop] = entity[prop]
+                return target[prop]
+            }
+            throw new Error(`Property ${prop} does not exist on entity '${fq}' or cds.entities is not yet defined`)
+        },
+        set: function (target, prop, newVal) {
+            target[prop] = newVal
+            return true
```
daogrady, line 26:
```suggestion
```
looks like the setter is currently just the default behaviour

daogrady:
Hi Ludwig,

thanks for checking in on those two points!
I ran similar benchmarks but mocked the actual call to `cds.entities`, which left me at a performance hit of around factor 20. It doubled to factor 40 when I forced repeated CSN lookups by deliberately causing cache misses during property access:

```js
for (let i = 0; i < 1000000; i++) {
  MyEntity[i];  // look for another property in every iteration
}
```

your results look even more daunting, as access times go up by _two orders of magnitude_, even for the cached case. While each individual property access is still negligible (0.00028ms per access in your results), this is in the ballpark of where I will have to check back with higher-ups, as this would mean using cds-typer would stray further away from being a purely design-time-thing, and would come with additional runtime implications.

So I am afraid I have to postpone this feature yet again until we get the chance to bring this up in the appropriate meetings. But I am hopeful that we can get this to at least be an opt-in.

Best,
Daniel

stockbal:
Hi Daniel,

understood. Just wanted to check if you tested also with the latest approach  that I mentioned here at the end <https://github.com/cap-js/cds-typer/pull/281#issuecomment-2325148356>.

| Entity Origin   | Test Scenario (1M iterations)             | Duration in ms | Factor |
| --------------- | ----------------------------------------- | -------------- | ------ |
| Proxy           | New Property access in each iteration     | ~119           | ~12     |
| `cds.entities` | New Property access in each iteration     | ~10            |        |
| Proxy           | Access of `elements` property in each iteration | ~25            | ~20     |
| `cds.entities` | Access of `elements` property in each iteration | ~1             |        |

It's weird that for the repeated access to the same property the factor is higher than for random property access.

Regards,
Ludwig

daogrady:
Hi Ludwig,

no, I missed that. 🙂 That does look a lot better and should also significantly reduce space complexity, as your new approach does not build an effective clone of every entity in the CSN. 👍
I will give it a more thorough look tomorrow.

Best,
Daniel

daogrady:
Looks really good! :)
I managed to squeeze out a few more milliseconds by removing the check for `initialized` on every property access by modifying the getter, and by eliminating the need to go up the prototype chain via indirection. But we are now in the realm of micro optimisations.

```js
const proxyAccessFunction = function (namespace, name, opts = {}) {
    const fq = namespace !== '' ? `${namespace}.${name}` : name
    return new Proxy(
        { ...opts, name: fq },
        {
            get: function (_, prop) {
                if (!cds.entities) throw Error('cds.entities not available yet')
                const entity = cds.entities[fq]
                this.get = (_, prop) => entity[prop]
                return entity[prop]
            }
        }
    )
}
```

stockbal:
Hi Daniel,

I like the override idea of the getter. But your new approach has some flaws when it comes to inline enum definitions

```js
module.exports.Book.Genre ??= { Action: "Action", Science_Fiction: "Science-Fiction" }
```

An import of such a file would immediately throw the exception as the proxy is trying to access `Genre`.  And we also loose access to the enums if you only return properties from the `cds.entities` reference.

Another point that I don't like now about the exception, you are no longer able to access the `name` of the entity which is already determined and `cds.entities` is not required to access it.

We could make a combination of my 2 versions of and the getter-override you introduced:

```js
function (namespace, name, opts = {}) {
    const { target, customProps } = { target: {}, customProps: [], ...opts }
    const fq = namespace !== '' ? `${namespace}.${name}` : name
    let entity
    return new Proxy(target, {
        get: function (target, prop) {
            if (cds.entities) {
                target.__proto__ = cds.entities[fq]
                this.get = (target, prop) => target[prop]
                return target[prop]
            }
            // we already know the name so we skip the cds.entities proxy access
            if (prop === 'name') return fq
            // custom properties access on 'target' as well as cached _entity property access goes here
            if (Object.hasOwn(target, prop)) return target[prop]
            // inline enums have to be caught here for first time access, as they do not exist on the entity
            if (customProps.includes(prop)) return target[prop]
            // last but not least we pass the property access to cds.entities
            throw new Error(`Property ${prop} does not exist on entity '${fq}' or cds.entities is not yet defined`)
        }
    })
}
```

Until `cds.entities` is defined we can then still access the `name` and the enums without getting an error. As soon as `cds.entities` exists and the proxy is accessed the getter override will take place.

Regards,
Ludwig

daogrady:
Hi Ludwig,

good point! I am a bit concerned that users would be a bit confused why accessing _some_ properties in _some_ places causes an error, while others don't. But overall, this looks pretty solid and I am rather confident we can justify this change. I will bring this to the next meeting so we can hopefully ship this feature soon.
I'd still expect this to be an opt-in at first, so we should be prepared to make the wrapping conditional.

Best,
Daniel

stockbal:
Hi Daniel,

great. When you speak about opt-in I am assuming you mean a CLI option like:

```shell
--useEntitiesProxy: <true | false>
(default: false)

If set to true the `cds.entities` exports in the generated `index.js` files will be wrapped in `Proxy` objects
so static import/require calls can be used everywhere.

WARNING: entity properties can still only be accessed after `cds.entities` has beed loaded
```

And if the setting is false the classic export approach is used.
Should I already add this to the PR?

Regards,
Ludwig.

daogrady:
Hi Ludwig,

yes, that is what I meant. 🙂 If you could include that in the PR, then that would surely speed things up. Let me know if you need help getting that in or if anything is unclear.

Best,
Daniel

stockbal:
Hi Daniel,

I added the CLI option as requested to enable the proxies for entities only when requested.

Regards,
Ludwig

daogrady (approved):
Hi Ludwig,

I checked the changes and ran some more benchmarks. Runtime looks totally fine now.
We also discussed this change yesterday and agreed that this is a valuable addition, so thanks for putting in this tremendous effort!
Let's get this merged and released 🚀

Best,
Daniel

PR #291, opened 2024-08-07
Signatures for unbound actions + named call style

chgeo:
Default export for the `<Service>.ts` file is now a service class:

```ts
import cds from '@sap/cds'
export default class extends cds.Service { ... }
```

Which declares functions for each unbound action, reusing the existing type:
```ts
export default class extends cds.Service {
  declare submitOrder: typeof submitOrder
}
```

The action type gets a second signature for the 'named/object' call style supported by Node.js:

```ts
export declare const sumitOrder: {
// positional
(book: __.DeepRequired<Book>['ID'], quantity: number | null): Promise< {
  stock?: number | null,
} | null | null> |  {
  stock?: number | null,
} | null | null
// named
({book, quantity}: {book: __.DeepRequired<Book>['ID'], quantity: number | null}): Promise< {
  stock?: number | null,
} | null | null> |  {
  stock?: number | null,
} | null | null, 
// metadata (do not use)
__parameters: {book: __.DeepRequired<Book>['ID'], quantity: number | null}, __returns: Promise< {
  stock?: number | null,
} | null | null> |  {
  stock?: number | null,
} | null | null, kind: 'action'};

```

So that in the end this works:

```ts
import CatalogService from '#cds-models/CatalogService'
const cat = await cds.connect.to(CatalogService)
await cat.submitOrder(book, quantity)
await cat.submitOrder({ book, quantity })
```

Required: 
- `@cap-js/cds-types` >= 0.6.4 (incl. https://github.com/cap-js/cds-types/pull/173)
- `@sap/cds` >= 8.2.0, allowing `cds.connect.to ({ name: 'service' })`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## Version 0.26.0 - TBD
+### Added
+- Apps need to provide `@sap/cds` version `8.2` or higher.
+- Apps need to provide `@cap-js/cds-types` version `0.6.4` or higher.
+- Typed methods are now generated for calls of unbound actions. Named and positional call styles are supported, e.g. `service.action({one, two})` and `service.action(one, two)`.
 
 ## Version 0.25.0 - 2024-08-13
 ### Added
 - Declaring a type alias on an enum in cds now also exports it on value level in the resulting type
-- Apps need to provide `@sap/cds` version `8.2` or higher.
-- Apps need to provide `@cap-js/cds-types` version `0.6.4` or higher.
-- Typed methods are now generated for calls of unbound actions. Named and positional call styles are supported, e.g. `service.action({one, two})` and `service.action(one, two)`.
 
 ### Fixed
 - Classes representing views and projections will no longer carry ancestry to avoid clashes thereof with aliases fields
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ## Version 0.25.0 - TBD
 ### Added
 - Declaring a type alias on an enum in cds now also exports it on value level in the resulting type
+- Apps need to provide `@sap/cds` version `8.2` or higher.
+- Apps need to provide `@cap-js/cds-types` version `0.6.4` or higher.
 
 ### Fixed
 - Classes representing views and projections will no longer carry ancestry to avoid clashes thereof with aliases fields
@@ -83,7 +85,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 - Importing an enum into a service will now generate an alias to the original enum, instead of incorrectly duplicating the definition
 - Returning entities from actions/ functions and using them as parameters will now properly use the singular inflection instead of returning an array thereof
 - Aspects are now consistently named and called in their singular form
-- Only detect inflection clash if singular and plural share the same namespace. This also no longer reports `sap.common` as erroneous during type creation 
+- Only detect inflection clash if singular and plural share the same namespace. This also no longer reports `sap.common` as erroneous during type creation
 
 ## Version 0.19.0 - 2024-03-28
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 - Declaring a type alias on an enum in cds now also exports it on value level in the resulting type
 - Apps need to provide `@sap/cds` version `8.2` or higher.
 - Apps need to provide `@cap-js/cds-types` version `0.6.4` or higher.
+- Typed methods are now generated for calls of unbound actions. Named and positional call styles are supported, e.g. `service.action({one, two})` and `service.action(one, two)`.
 
 ### Fixed
 - Classes representing views and projections will no longer carry ancestry to avoid clashes thereof with aliases fields
--- a/lib/file.js
+++ b/lib/file.js
@@ -169,7 +169,7 @@ class SourceFile extends File {
             : empty
         const callableSignatures = []
         if (callStyles.positional) {
-            callableSignatures.push(`\n// positional\n(${parameterTypes}): ${returns}`)
+            callableSignatures.push(`// positional\n(${parameterTypes}): ${returns}`)
         }
         if (callStyles.named) {
             const parameterNames = createObjectOf(parameters.map(([n]) => normalise(n)).join(', '))
@@ -182,7 +182,7 @@ class SourceFile extends File {
         }
         const kindDef = kind ? `, kind: '${kind}'` : ''
         const suffix = initialiser ? ` = ${initialiser}` : ''
-        const lambda = `{ ${callableSignatures.join('\n')}, \n// metadata (do not use)\n__parameters: ${parameterTypeAsObject}, __returns: ${returns}${kindDef}}`
+        const lambda = `{\n${callableSignatures.join('\n')}, \n// metadata (do not use)\n__parameters: ${parameterTypeAsObject}, __returns: ${returns}${kindDef}}`
         return prefix + lambda + suffix
     }
 
--- a/lib/file.js
+++ b/lib/file.js
@@ -175,7 +175,7 @@ class SourceFile extends File {
             const parameterNames = createObjectOf(parameters.map(([n]) => normalise(n)).join(', '))
             callableSignatures.push(`// named\n(${parameterNames}: ${parameterTypeAsObject}): ${returns}`)
         }
-        if (callableSignatures.length === 0)  throw new Error('At least one call style must be specified')
+        if (callableSignatures.length === 0) throw new Error('At least one call style must be specified')
         let prefix = name ? `${normalise(name)}: `: ''
         if (prefix && isStatic) {
             prefix = `static ${prefix}`
--- a/lib/file.js
+++ b/lib/file.js
@@ -141,6 +141,7 @@ class SourceFile extends File {
      * @param {'action' | 'function'} options.kind - kind of the lambda
      * @param {string} [options.initialiser] - the initialiser expression
      * @param {boolean} [options.isStatic] - whether the lambda is static
+     * @param {{positional?: boolean, named?: boolean}} [options.callStyles] - whether to generate positional and/or named call styles
      * @returns {string} the stringified lambda
      * @example
      * ```js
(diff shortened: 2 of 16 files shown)


chgeo:
@ogrady first draft, tests still failing

test/tscheck.js
```diff
@@ -29,7 +29,7 @@ function checkProgram (program) {
  * @param {object} opts - the options to pass to the TS compiler
  */
 async function checkTranspilation (apiFiles, opts = {}) {
-    const options = {...{ noEmit: true }, ...opts}
+    const options = {...{ noEmit: true, esModuleInterop: true }, ...opts}
```
chgeo, line 32: This resolves the _no default export_ test issue.

lib/file.js
```diff
@@ -161,19 +162,27 @@ class SourceFile extends File {
...
-    static stringifyLambda({name, parameters=[], returns='any', initialiser, isStatic=false, kind}) {
+    static stringifyLambda({name, parameters=[], returns='any', kind, initialiser, isStatic=false, callStyles={positional:true, named:true}}) {
         const parameterTypes = parameters.map(([n, t]) => `${normalise(n)}: ${t}`).join(', ')
         const parameterTypeAsObject = parameterTypes.length
             ? createObjectOf(parameterTypes)
             : empty
-        const callableSignature = `(${parameterTypes}): ${returns}`
+        const callableSignatures = []
+        if (callStyles.positional) {
+            callableSignatures.push(`\n// positional\n(${parameterTypes}): ${returns}`)
```
daogrady, line 172: I guess we don't need the leading `\n` here, or else we'd also have to add it [here](https://github.com/cap-js/cds-typer/pull/291/files#diff-387904f6b07263914bc1b0cb291a27832cc6963e01b5d44b75460771fc25ee8fR176) for `callStyles: {positions: false, named: true}`?

```suggestion
            callableSignatures.push(`// positional\n(${parameterTypes}): ${returns}`)
```

lib/file.js
```diff
@@ -161,19 +162,27 @@ class SourceFile extends File {
...
+        }
+        if (callStyles.named) {
+            const parameterNames = createObjectOf(parameters.map(([n]) => normalise(n)).join(', '))
+            callableSignatures.push(`// named\n(${parameterNames}: ${parameterTypeAsObject}): ${returns}`)
+        }
+        if (callableSignatures.length === 0)  throw new Error('At least one call style must be specified')
```
daogrady, line 178:
```suggestion
        if (callableSignatures.length === 0)  throw new Error('At least one call style must be specified')
        if (callableSignatures.length === 0) throw new Error('At least one call style must be specified')
```

lib/file.js
```diff
@@ -356,11 +365,13 @@ class SourceFile extends File {
      */
     getImports() {
         const buffer = new Buffer()
+        buffer.add('import cds from \'@sap/cds\'') // TODO should go to visitor#printService, but can't express this as Path
```
daogrady, line 368: There is in fact [one other place](https://github.com/cap-js/cds-typer/blob/main/lib/resolution/builtin.js#L10) where we refer to types from another lib. To circumvent top level imports I had resorted to inline-imports until now. But maybe this is a good occasion to facilitate top level imports in generated files. I am fine with either way, just adding this for completeness sake.

lib/visitor.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 const util = require('./util')
+const cds = require('@sap/cds')
```
daogrady, line 4:
```suggestion
```

lib/visitor.js
```diff
@@ -443,15 +444,40 @@ class Visitor {
...
+     * @param {string|undefined} doc - the documentation to add
+     * @param {import('./file.js').Buffer} buffer - the buffer to add the documentation to
+     */
+    #renderDoc(doc, buffer = new Buffer()) {
+        if (!doc) return ''
+        buffer.add('/**')
+        doc.split(/\r?\n/).forEach(line => { buffer.add(` * ${line}`) })
+        buffer.add(' */')
+        return buffer.join()
+    }
```
daogrady, line 479: This looks similar to what [`docify`](https://github.com/cap-js/cds-typer/blob/main/lib/components/wrappers.js#L62) does. Can that be reused?
Also doesn't look like the `return` is required, as you're manipulating the passed `buffer`. I guess the argument should not be optional either.

package.json
```diff
@@ -40,10 +40,13 @@
   "bin": {
     "cds-typer": "./lib/cli.js"
   },
-  "dependencies": {
-    "@sap/cds": ">=7.7"
+  "peerDependencies": {
```
daogrady, line 43: I can not come up with any scenario on the spot, but are there use cases where there is no `@sap/cds` directly available? CI/CD? Within the vscode extension? 🤔

test/smoke/setup.js
```diff
@@ -1,4 +1,4 @@
-const { fs } = require('@sap/cds/lib/utils/cds-utils')
```
daogrady, line 1: 😬😗🎶

test/util.js
```diff
@@ -138,6 +139,19 @@ const cds2ts = async (cdsFile, options = {}) => typer.compileFromFile(
...
+        devDependencies: {
+            '@sap/cds': 'file:' + require('@sap/cds').home,
+            '@cap-js/cds-types': 'file:' + path.resolve(require.resolve('@cap-js/cds-types/package.json'), '..')
+        }
+    }, null, 2))
+    // Avoid installing dependencies if they are already present. Speeds up tests quite a bit.
+    if (!fs.existsSync(path.join(projDir, 'node_modules'))) {
+        execSync('npm install', { cwd: projDir })
+    }
+}
```
daogrady, line 153: nifty!

test/util.js
```diff
@@ -166,9 +180,12 @@ async function prepareUnitTest(model, outputDirectory, parameters = {}) {
 
     if (parameters.transpilationCheck) {
         const tsFiles = paths.map(p => path.join(p, 'index.ts'))
+        // create a package.json w/ common dependencies in a higher dir so that they can be reused by many tests
+        createProject(path.resolve(outputDirectory, '../..'))
```
daogrady, line 184: without having checked, this reads a bit like it could cause rogue files in our monorepo setup. Did you perchance check that?

lib/visitor.js
daogrady, line 479:
```suggestion
    #renderDoc(doc, buffer) {
        docify(doc).forEach(d => buffer.add(d))
    }
```
I guess?
Or just inline it at the appropriate questions

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ## Version 0.25.0 - TBD
 ### Added
 - Declaring a type alias on an enum in cds now also exports it on value level in the resulting type
+- Apps need to provide `@sap/cds` version `8.2` or higher.
+- Apps need to provide `@cap-js/cds-types` version `0.6.4` or higher.
```
daogrady, line 11: I guess the feature itself also deserves an entry here 😃

lib/visitor.js
```diff
@@ -443,15 +444,40 @@ class Visitor {
...
         const { namespace } = this.entityRepository.getByFq(fq)
         const file = this.fileRepository.getNamespaceFile(namespace)
-        // service.name is clean of namespace
-        file.services.buffer.add(`export default { name: '${service.name}' }`)
+        const buffer = file.services.buffer
+
+        this.#renderDoc(service.doc, buffer)
+        // file.addImport(new Path(['cds'], '')) TODO make sap/cds import work
+        buffer.addIndentedBlock('export default class extends cds.Service {', () => {
+            Object.entries(service.operations||{}).forEach(([name, {doc}]) => {
```
daogrady, line 461:
```suggestion
            Object.entries(service.operations ?? {}).forEach(([name, {doc}]) => {
```

daogrady (changes requested):
Thanks for the addition! 
Suggestions are mainly pebbles, except for maybe the changelog entry (and possibly the potential refactor of `#renderDoc`).

lib/file.js
chgeo: I guess I have added it to have a line break after the `{`:
```ts
export declare const submitOrder: { 
// positional
```

Didn't know how to this better.

package.json
chgeo: > Within the vscode extension?

Good point.  Is cds-typer installed w/ our extension?  Don't we rely on the app installing it?  In this case, we are fine.

test/util.js
chgeo: From what I have seen, the tests are always executed in tmpdir, so to rouge source files.

lib/file.js
chgeo: Never mind, I have moved the line break to a later place.

CHANGELOG.md
chgeo: Done

lib/file.js
chgeo: Let's leave it like that for now, to not make this PR grow too much.

lib/visitor.js
```diff
@@ -389,10 +389,12 @@ class Visitor {
         const returnType = operation.returns
             ? this.resolver.resolveAndRequire(operation.returns, file)
             : { typeName: 'void', typeInfo: { isArray: false, inflection: { singular: 'void', plural: 'void' } } }
-        const returns = this.inlineDeclarationResolver.getPropertyDatatype(
+        let returns = this.inlineDeclarationResolver.getPropertyDatatype(
             returnType,
             returnType.typeInfo.isArray ? returnType.typeName : returnType.typeInfo.inflection.singular
         )
+        // results of operations are effectively a Promise at runtime
+        returns = operation.returns ? `Promise<${returns}>` : returns
```
chgeo, line 397: @daogrady To make operations' return types work, I had to wrap the result type into a `Promise`.  This seems to reflects the runtime behavior.  They always return promises, even if their impl. is non-async.  The [docs](https://cap.cloud.sap/docs/guides/providing-services#calling-actions-functions) indicate the same, although we might want to cross-check w/ the runtime folks.

chgeo: @sjvans maybe you can comment on return values of actions/functions. Always Promises?

daogrady: Sure! I have moved the promisification to a [module with similar functions](https://github.com/cap-js/cds-typer/blob/main/lib/components/wrappers.js). Did you check if that had any implications on handler implementation? I.e.


```cds
// MyService.cds
namespace my.namespace;

service MyService {
    action myUnboundAction returns Integer;
}
```

```js
// MyService.js|ts
const { myUnboundAction } = require('#cds-models/MyService')

// ...

this.on(myUnboundAction, () => {
  return 42   // will the type system now bug the user about 42 not being a Promise?
})
```

chgeo: Good point.  Indeed, promises are only accepted if the handler is async.  If it is sync, it gets rejected - at least it should be rejected, but is 'catched' by a different `on` overload that has an `any` return type.  A topic for a different day, I'd say.

I have changed the type to `Promise<return> | return`.

daogrady (approved):
looks good now 🙂

chgeo:
As discussed, wait with the merge til after the next release.

PR #293, opened 2024-08-13
Add prerelease workflow

daogrady:
We currently have a rather unreliable release cycle and sometimes wait for several features to be completed before we bundle them into one release. This can be cumbersome to users who wait for a particular feature to be released.
This PR therefore introduces a workflow ~~that automatically pushes a new `-beta` version to npmjs whenever changes land in `main`. Prereleases can be skipped by adding `[skip prerelease]` to the merge commit head, to avoid flooding users with new betas for every tiny change.~~ to conveniently release a `beta` version to npmjs when triggered manually.

Caveat: if we decide to employ this workflow we would have to always bump the package version to the next expected number right _after_ a release. We currently only bump the version number right _before_ a release. But since the version number for the prereleases are derived from the current version found in package.json, that would create a beta version for the last released version, which doesn't make sense.


.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          BRANCH_NAME="prerelease-${{ steps.bump_version.outputs.prerelease_version }}"
+          git checkout -b $BRANCH_NAME
+          git push origin $BRANCH_NAME
+
+      - name: Push changes
+        run: |
+          git add package.json package-lock.json
+          git commit -m "Bump version to ${{ steps.bump_version.outputs.prerelease_version }}"
+          git push origin $BRANCH_NAME
```
chgeo, line 45: Could we also create and push a tag pointing to this version?

chgeo (commented):
The [npm docs](https://docs.npmjs.com/cli/v10/commands/npm-version#description) give some more advice on how to structure the workflow.  We'd need less Git interactions then.  Have you checked this?

chgeo (approved):

PR #296, opened 2024-08-15
Support optional operation parameters

chgeo:
If a `@Core.OptionalParameter` annotation is set on an operation parameter, mark the property as optional in the named call style.
The actual annotation is not fixed yet, but this is the state of discussion.

Resulting `.ts` file with `?` modifiers:
```ts
export declare const submitOrder: {
  // positional
  (book: ..., quantity: number | null): ...
  // named
  ({book, quantity}: {book: ..., quantity?: number | null}): ...
  ...
}
```

The existing default for mandatory/optional is not changed, so no migration expected.

Continuation of https://github.com/cap-js/cds-typer/pull/291

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -146,7 +146,7 @@ class SourceFile extends File {
      * Stringifies a lambda expression.
      * @param {object} options - options
      * @param {string} options.name - name of the lambda
-     * @param {[string, string][]} [options.parameters] - list of parameters, passed as [name, type] pairs
+     * @param {[string, string, string][]} [options.parameters] - list of parameters, passed as [name, modifier, type] pairs
      * @param {string} [options.returns] - the return type of the function
      * @param {'action' | 'function'} options.kind - kind of the lambda
      * @param {string} [options.initialiser] - the initialiser expression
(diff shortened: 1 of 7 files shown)


chgeo:
Check for pos. call style whether we can support `?`

lib/file.js
```diff
@@ -169,17 +169,18 @@ class SourceFile extends File {
...
     static stringifyLambda({name, parameters=[], returns='any', kind, initialiser, isStatic=false, callStyles={positional:true, named:true}}) {
-        const parameterTypes = parameters.map(([n, t]) => `${normalise(n)}: ${t}`).join(', ')
+        const parameterTypes = parameters.map(([n, mod='', t]) => `${normalise(n)}${mod}: ${t}`).join(', ')
         const parameterTypeAsObject = parameterTypes.length
             ? createObjectOf(parameterTypes)
             : empty
         const callableSignatures = []
         if (callStyles.positional) {
-            callableSignatures.push(`// positional\n(${parameterTypes}): ${returns}`)
+            const paramTypesPositional = parameters.map(([n, , t]) => `${normalise(n)}: ${t}`).join(', ') // must not include ? modifiers
```
daogrady, line 182: depending on how badly we want optional positional parameters:
```suggestion
            // the first optional parameter makes all succeeding params optional as well
            let optional = parameters.at(0)?.[1] === '?'
            const paramTypesPositional = parameters.map(([n, mod, t]) => {
                optional ||= mod === '?'
                return `${normalise(n)}${optional ? '?' : ''}: ${t}`
            }).join(', ')
```

daogrady:
Thanks for the contribution! Looks good to me. I have added a suggestion for enabling optional positional parameters, but I have no strong opinion on whether to include it or not. You probably have a better overview if this would be useful.

Do we need another changelog entry or is the contribution sufficiently described by the one in https://github.com/cap-js/cds-typer/pull/291 ?

lib/file.js
chgeo: I don't think we need this dynamic behavior just to avoid one parameter.
The current approach of offering the optionals in one call style only is more straight forward.

chgeo:
> Do we need another changelog entry or is the contribution sufficiently described by the one in #291 ?

Added one.

PR #297, opened 2024-08-19
Tests: Generate, TS check and run

chgeo:
New test to specify assertions in handler code from a consumer perspective.
Then, it
- runs typer
- runs TSC to check/transpile
- calls `cds.serve`, thereby loading handlers
- [later] invokes services to execute e.g. actions


test/unit/tsCheckAndRun.test.js
```diff
@@ -0,0 +1,47 @@
...
+        // 'inflection',
+        // 'inheritance',
+        // 'inline',
+        // 'notnull',
+        // 'projection',
+        // 'references',
+        // 'scoped',
+        // 'type',
+        // 'typeof',
+        // 'views',
```
daogrady, line 32: I see that you are currently just testing the waters with `actions`, but still wanted to note that this can be [resolved dynamically](https://github.com/cap-js/cds-typer/blob/main/test/smoke/transpilation.test.js#L8) for all existing model directories

chgeo: Done

PR #299, opened 2024-08-21
Add extra export for services

chgeo:
In CommonJS modules, our default exports for services don't work.

Ok:
image

Wrong:
image

Not sure this is a bug or works as intended.

Fix is to create a separate export.

Keep the current default export as it's already used. This is what a search in SAP's GH showed.


chgeo:
Prefer named export, esp. in docu

daogrady (approved):
👍 Thanks!

PR #300, opened 2024-08-26
Fix doc comments for operation parameters

chgeo:
Make sure comments for operations and their params show up properly for
- implementors, i.e. `.on(action,...)`
- callers, i.e. `action({param})`

Also make doc generation better:
- don't break for one-line doc comments, e.g. `/** one line */`
- escape illegal `*/` sequences

Depends on https://github.com/cap-js/cds-typer/pull/299


daogrady (approved):
LGTM, thanks!

PR #308, opened 2024-09-04
Generate static .keys property

### LINKED ISSUE #294: [FEATURE] Expose 'keys' & 'gens' as static props per each compiled CDS entity
### Description

I need to know which CDS property names are marked as CDS keys & which are auto-generated if to not provide them explicitly.

### Suggested Solution

CDS entity with primary key & auto-generated props:

```cds
entity TestObjects: cuid {
  name: String not null;
  description: String;
  type: String not null default 'test';
  updatedAt: Timestamp not null default $now;
}
```

It has to be compiled into:

```TypeScript
class TestObjects {
  ...
  ID: string;
  name: string;
  description: string | null;
  type: string;
  updatedAt: String;
  ...
  /** primary keys of CDS entity */
  static readonly keys: ("ID")[];
  /** auto-generated properties of CDS entity */
  static readonly gens: ("ID" | "type" | "updatedAt")[];
}
```

### Alternatives

_No response_

### Additional Context

_No response_

daogrady:
Fixes #294

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 
 ## Version 0.22.0 - 2024-06-20
+
+### Added
+- Added a static `.keys` property in all entities. That property is dictionary which holds all properties as keys that are marked as `key` in CDS
+
 ### Fixed
 - Fixed a bug where keys would sometimes inconsistently become nullable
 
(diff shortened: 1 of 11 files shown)


daogrady:
Hi @siarhei-murkou,

I have adjusted the `.keys` property as discussed to use an internal `Symbol` as key to make it unaccessible inside of handlers.
You can get an idea of which usage is allowed and which is not from [the associated test file](https://github.com/cap-js/cds-typer/blob/feat/static-keys-property/test/unit/files/keys/model.ts).
Please let me know if this covers the cases we were discussing.

Best,
Daniel

PR #311, opened 2024-09-04
Disable positional call style for ABAP RFC actions

chgeo:
Actions for ABAP RFC modules should only be callable with named parameters. Such modules have 'parameter categories' (import/export/changing/tables) that cannot be called in a flat order.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -218,7 +218,7 @@ class SourceFile extends File {
      * @param {import('./typedefs').visitor.ParamInfo[]} parameters - list of parameters, passed as [name, modifier, type] tuple
      * @param {string} returns - the return type of the function
      * @param {'function' | 'action'} kind - kind of the node
-     * @param {?string[]} doc - documentation for the function
+     * @param {string[]} doc - documentation for the function
      * @param {{positional?: boolean, named?: boolean}} callStyles - how the operation can be called
      */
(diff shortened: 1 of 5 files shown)


lib/file.js
```diff
@@ -218,12 +218,13 @@ class SourceFile extends File {
      * @param {import('./typedefs').visitor.ParamInfo[]} parameters - list of parameters, passed as [name, modifier, type] tuple
      * @param {string} returns - the return type of the function
      * @param {'function' | 'action'} kind - kind of the node
-     * @param {string[]?} doc - documentation for the function
+     * @param {?string[]} doc - documentation for the function
```
daogrady, line 221: I am unfamiliar with the leading `?` in parameter types. Does that have special semantics or is it a typo?

chgeo: typo :)

chgeo: Fixed

chgeo:
> LGTM (here's to hope we don't have to support _too_ many special forms, endpoints, quirks beyond RFC)

Yes, I hope so too.
ABAP is always special :)

chgeo:
But we can also discuss w/ the importer/compiler colleagues if they can give us an annotation hint (`@cds.action.callstyle.positional:false` or so).

PR #315, opened 2024-09-06
Fix CDS prop nullability if `@mandatory` is used

siarhei-murkou:
Hi @daogrady 👋 ,

Since the version `8.0.2`, CDS has introduced the new logic for CDS params in CDS functions/actions. Now, we have to use `@mandatory` annotation instead of setting `not null` keyword. Whereas `not null` is still working on CDS entity layer.. 
Therefore, I have added a check on `@mandatory` existence in order to determine if the property should/shouldn't be marked as nullable in the compiled TypeScript files.

### DIFF
--- a/lib/resolution/resolver.js
+++ b/lib/resolution/resolver.js
@@ -452,7 +452,7 @@ class Resolver {
             isArray: false,
             isNotNull: element?.isRefNotNull !== undefined
                 ? element?.isRefNotNull
-                : element?.key || element?.notNull || cardinality > 1,
+                : element?.key || element?.notNull || element?.['@mandatory'] || cardinality > 1,
         }
 
         if (element?.type === undefined) {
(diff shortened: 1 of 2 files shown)


daogrady:
Hi Siarhei,

thanks for providing this addition! Looks good on first glance, but there seem to be no tests covering this new behaviour as of now. Can you please add a test checking a mandatory vs a non-mandatory field? A [more approachable way of testing](https://github.com/cap-js/cds-typer/wiki/Testing#-compilation-tests) has been added recently, where you basically have to write consuming TS code for your generated files. Let me know when you need help writing this test.

Best,
Daniel

daogrady:
Also, as I have learned today, this feature may still slightly change in the near future. So let's maybe shelf this for now until we know with certainty how `@mandatory` and `not null` will turn out.

siarhei-murkou:
> Also, as I have learned today, this feature may still slightly change in the near future. So let's maybe shelf this for now until we know with certainty how `@mandatory` and `not null` will turn out.

Hi Daniel, sorry for late response! 😞 

Okay, I see.. Lets then wait for the final decision by _cds runtime_ team. 🙂 
I suppose, right now this PR might be closed...

So far, I'm using `@mandatory` together with `not null` as a workaround.

PR #316, opened 2024-09-09
Add docs for enums, fix it for entities

chgeo:
Add docs to exposed artifacts so that they appear in code completion etc.

- for enums: on the exported const and the element pointing to it
- for entities: on the exported class, not the aspect


daogrady (approved):
Thanks!

PR #324, opened 2024-09-12
Snake case options

daogrady:
Configuration parameters can now be passed either in camelCase or snake_case. This mechanism works for both CLI parameters, as well as for options passed via `cds.env`. The order of precedence is (still):

CLI parameters > `cds.env` > default values

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -38,7 +38,7 @@ class Config {
     }
 
     constructor() {
-        this.values = {...Config.#defaults, ...(cds?.env?.typer ?? {})}
+        this.values = {...Config.#defaults, ...(cds.env.typer ?? {})}
         this.proxy = camelSnakeHybrid(this.values)
         // proxy around config still allows arbitrary property access:
         // require('config').configuration.logLevel = 'warn' will work
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,6 +1,11 @@
 const { camelToSnake } = require('./util')
 
 /**
+ * @typedef {import('./typedefs').config.cli.EnrichedParameter} EnrichedParameter
+ * @typedef {import('./typedefs').config.cli.Parameter} Parameter
+ */
+
+/**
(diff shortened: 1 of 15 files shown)


lib/config.js
```diff
@@ -0,0 +1,101 @@
...
+}
+class Config {
+    static #defaults = {
+        propertiesOptional: true,
+        useEntitiesProxy: false,
+        inlineDeclarations: 'flat'
+    }
+
+    constructor() {
+        this.values = {...Config.#defaults, ...(cds?.env?.typer ?? {})}
```
chgeo, line 41: `cds.env` should never be null/undefined.

```suggestion
        this.values = {...Config.#defaults, ...(cds.env.typer ?? {})}
```

Also, `cds.env` uses `process.cwd()`, so are we sure we run in the project root?
Otherwise we'd need to use `cds.env.for('cds', <projectRoot>)`

chgeo: > there is no guarantee cds-typer is started from the project root. How would we find `<projectRoot>`?

That's a classic question for which we don't have a good answer.  Often this is done by finding a package.json w/ sap/cds as a dependency, or a .cdsrc.json.
But before we go there, I wonder why this hasn't been an issue so far.  How is the location determined for typer in VS Code?  Otherwise `cds-typer '*'` would not work.

In CIs I guess the project location is correctly set.

chgeo: Yes, we should settle on a standard way to find a project root given a file.
But let's do that in a separate PR.

> we apply a naive heuristic going up the a file structure until we find a directory that looks like a root directory.

I thought there is already an implementation in cds-typer to do that which we could use.  But I guess this is vscode-cds then?

daogrady: > I thought there is already an implementation in cds-typer to do that which we could use. But I guess this is vscode-cds then?

Correct, this is currently part of cds-lsp.

chgeo (approved):

PR #332, opened 2024-09-18
Determine optional action parameters from CSN's `notNull`

chgeo:
Replace the use of `@Core.OptionalParameter` for optional parameters (which was an intermediate solution anyways) by evaluating CSN's `notNull` property.  This is in line with the decision in the CDS spec meeting and with what RFC importer now generates.

Note that this formally inverses the default for optional parameters:
- They are all optional by default, which is how the runtime treats it.
- Mandatory parameters now have to be marked with `notNull` (or `not null` in CDL).

### DIFF
--- a/lib/resolution/resolver.js
+++ b/lib/resolution/resolver.js
@@ -63,8 +63,7 @@ class Resolver {
      * @returns {boolean} whether the type is configured to be optional
      */
     isOptional(type) {
-        // TODO temporary solution to determine optional parameters. Align w/ compiler/importer.
-        return Object.keys(type).some(k => k.startsWith('@Core.OptionalParameter'))
+        return !type.notNull
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## Version 0.27.0 - TBD
+### Changed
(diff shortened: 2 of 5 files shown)


daogrady:
related: https://github.com/cap-js/cds-typer/pull/315

daogrady:
> by evaluating CSN's notNull property.

Has this property been present all along or does it require a certain version of CDS?

chgeo:
> Has this property been present all along or does it require a certain version of `@sap/cds`, the compiler, or any other related lib?

Has been there for a long time.

daogrady (approved):
thanks!

lib/resolution/resolver.js
```diff
@@ -63,8 +63,7 @@ class Resolver {
      * @returns {boolean} whether the type is configured to be optional
      */
     isOptional(type) {
-        // TODO temporary solution to determine optional parameters. Align w/ compiler/importer.
-        return Object.keys(type).some(k => k.startsWith('@Core.OptionalParameter'))
+        return !type.notNull
```
siarhei-murkou, line 66: Hi @daogrady & @chgeo 👋 🙂  

I found out a small bug in current `cds-typer@0.27.0` version that CDS array-like params are always compiled into optional TS props.

_A quick example to reproduce_:
```cds
function check(
    @mandatory param_1: String not null,
    @mandatory param_2: Array of String not null
) returns ...
```

_which leads to wrong compiled TS props_:
```TypeScript
export declare const check: {
    ...
    __parameters: { param_1: string; param_2?: string[] };
    ...
}
```

The hotfix is bellow:
```suggestion
        return type.items ? !type.items.notNull : !type.notNull
```

chgeo: Thanks for catching this @siarhei-murkou.  Will fix it.

Issue #323, opened 2024-09-12
[BUG] Missing option to pass CLI options to typer during 'cds build'


stockbal:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

I added the new cli option `--useEntitiesProxy true` to my VS Code settings (i.e. `"cds.typeGenerator.command"`) and during development the types are getting generated correctly.

However, during the build, cds-typer is using the function `compileFromFile` directly and only passes the output directory as option. This results in not having the proxy exports in the final cds build output.

### Expected Behavior

Some central way to specify cds-typer parameters that are used during development and during build, so the generated types output between dev and build does not differ.

### Environment

```markdown
| Module                 | Version |
| ---------------------- | --------|
| @cap-js/asyncapi       | 1.0.2   |
| @cap-js/cds-typer      | 0.26.0  |
| @cap-js/cds-types      | 0.2.0   |
| @cap-js/change-tracking| 1.0.7   |
| @cap-js/hana           | 1.2.0   |
| @cap-js/openapi        | 1.0.5   |
| @cap-js/sqlite         | 1.7.3   |
| @sap/cds               | 8.2.1   |
| @sap/cds-compiler      | 5.2.0   |
| @sap/cds-dk            | 8.2.0   |
| @sap/cds-dk (global)   | 7.9.6   |
| @sap/cds-fiori         | 1.2.7   |
| @sap/cds-foss          | 5.0.1   |
| @sap/cds-mtxs          | 2.1.0   |
| @sap/eslint-plugin-cds | 3.0.4   |
| Node.js                | v20.16.0|
```

daogrady:
Hi Ludwig,

thanks for pointing this out. I have some good news, then some bad news, then some more good news:
cds-typer can be fully configured using `cds.env`. More specifically, you can place any of the relevant options under `cds.env.typer` just as you would pass them in as parameters. Those should then be picked up by cds-typer, regardless of whether you run it from the IDE or as part of your build process. But as I have recently found out, camelCased environment options can not be used in all constellations for technical reasons. I am therefore currently [in the process](https://github.com/cap-js/cds-typer/pull/324) of allowing any option to be passed in both camelcase and snake_case.

Best,
Daniel

daogrady:
Hi Ludwig,

I have now merged the configuration via snake case, so it is available in `main` now if you would like to confirm this works.

Best,
Daniel

stockbal:
[...]

PR #334, opened 2024-09-19
Fix output path in build task

daogrady:
Fixes https://github.com/cap-js/cds-typer/issues/323

### DIFF
--- a/cds-plugin.js
+++ b/cds-plugin.js
@@ -36,7 +36,7 @@ const rmDirIfExists = dir => {
  * Remove files with given extensions from a directory recursively.
  * @param {string} dir - The directory to start from.
  * @param {string[]} exts - The extensions to remove.
- * @returns {Promise<void>}
+ * @returns {Promise<unknown>}
  */
 const rmFiles = async (dir, exts) => fs.existsSync(dir)
     ? Promise.all(
--- a/cds-plugin.js
+++ b/cds-plugin.js
@@ -68,12 +68,15 @@ cds.build?.register?.('typescript', class extends cds.build.Plugin {
 
     get #appFolder () { return cds?.env?.folders?.app ?? 'app' }
 
+    /**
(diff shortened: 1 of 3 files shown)


cds-plugin.js
```diff
@@ -89,7 +94,9 @@ cds.build?.register?.('typescript', class extends cds.build.Plugin {
 
     async #runCdsTyper () {
         DEBUG?.('running cds-typer')
-        await typer.compileFromFile('*', { outputDirectory: this.#modelDirectoryName })
+        cds.env.typer ??= {}
+        cds.env.typer.outputDirectory ??= this.#modelDirectoryName
```
stockbal, line 98: This does not have the expected effect, due to the fact that the module `./lib/compile` is implicitly loading module `./lib/config`.  So, the constructor of the `Config` class has already been executed and any changes to `cds.env.typer` will be ignored.

You would have to change it to

```js
const { configuration } = require('@cap-js/cds-typer/lib/config')
...
configuration.outputDirectory ??= this.#modelDirectoryName
```

daogrady: You're right, the properties should actually not be imprinted into the config at all, but dynamically queries from `cds.env`.
The joys of global state... I will check it again.

daogrady: (the fix should actually workd, as line 98 is basically a no-op, since `#modelDirectoryName` will attempt to read the `cds.typer.outputDirectory`, but it is still bad style)

daogrady: Ah, I see your use case now and I am afraid you are correct.
I have made the initialisation of `configuration` lazy to support modifications of `cds.env` before the first actual use of `cds-typer`.

stockbal: Looks good. The build runs through without any errors now.

daogrady: Great! Thanks for bringing this to our attention and for the review. 🙂

PR #344, opened 2024-10-03
feat: cds json schema for `cds.typer`

stockbal:
Adds schema definitions for `cds.typer` options to enable autocompletion and validation of typer options inside `package.json` and `.cdsrc*.json` files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## Version 0.28.0 - TBD
+### Added
+- Schema definition for `cds.typer` options in `package.json` and `.cdsrc-*.json` files
 
 ## Version 0.27.0 - 2024-10-02
 ### Changed
--- a/package.json
+++ b/package.json
@@ -60,5 +60,65 @@
       "test/smoke.jest.config.js",
       "test/unit.jest.config.js"
     ]
+  },
+  "cds": {
+    "schema": {
+      "cds": {
+        "typer": {
+          "type": "object",
+          "description": "Configuration for CDS Typer",
+          "properties": {
+            "inlineDeclarations": {
+              "type": "string",
+              "description": "Whether to resolve inline type declarations flat: (x_a, x_b, ...) or structured: (x: {a, b})",
+              "enum": [
+                "flat",
+                "structured"
+              ],
+              "default": "structured"
+            },
+            "IEEE754Compatible": {
+              "type": "boolean",
+              "description": "If set to true, floating point properties are generated as IEEE754 compatible '(number | string)' instead of 'number'",
+              "default": false
+            },
+            "jsConfigPath": {
+              "type": "string",
+              "description": "Path to where the jsconfig.json should be written.\nIf specified, cds-typer will create a jsconfig.json file and set it up to restrict property usage in types entities to existing properties only"
+            },
+            "logLevel": {
+              "type": "string",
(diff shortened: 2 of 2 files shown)


joergmann (approved):

daogrady:
Hi Ludwig,

thank you for this contribution! I hadn't even considered this so far and I think it makes a lot of sense and is very helpful.
I am just not very happy with effectively duplicating the options and descriptions in _cli.js_ and _package.json_. Plus, the ones you have listed so far do not include the optional snake_case syntax for the options.

I think it would make more sense in the long run to generate the schema in _package.json_ based on _cli.js_`.flags` to make sure it is always up to date without having to maintain it manually. One fitting course of action could be having an `npm` script that updates _package.json_ and, for the long run, having a CI script that automatically opens a PR when _package.json_ is out of sync with _cli.js_`.flags`. But that is pie in the sky and I'd be fine with merging your PR for now. The only thing I'd really like to be included are the snake_case flags. 👍

Best,
Daniel

stockbal:
Hi Daniel,

sounds all reasonable. Just a quick question about the snake-case options. You mentioned you added those because camel-case is not always usable. Can you tell my where and why?

Regards,
Ludwig

daogrady:
Hi Ludwig,

sure, camelCase variables apparently cause problems when configured via environment variables, so snake_case is the preferred style. Although I have yet to find a section on CAPire describing this, and the colleague who pointed this out is currently on vacation, so I'd have to pick this up next week to give you a definitive answer.

Best,
Daniel

daogrady:
Yes, you are right that it does not play a role technically. 🙂
We're trying to nudge users towards snake_case, as it is the more widely compatible use. We have also opted to only show snake_case in the [configuration examples on CAPire](https://cap.cloud.sap/docs/tools/cds-typer#configuration), even though camelCase is perfectly valid there too.
I guess the question is whether we are aiming for completeness or clarity here. If you'd prefer to only show one "set" of configuration options, we should use the snake_case versions rather than the camelCase ones.

Again: as a first version of this feature it's totally fine to just have them hard coded in _package.json_. We can follow up on dynamically updating _package.json_ later.

stockbal:
Ok. Then I would switch it to snake_case only.

Regarding the dynamic update. Would it be possible to totally keep the schema from the package.json in git and just add it via script call right before the `npm publish` step in the release workflow?

stockbal:
Hi Daniel,

I made the changes to snake_case. I also created a script to write the options from `/lib/cli.js#flags` to package.json. We just have to check if we can include this script in the release workflow, then we could remove the schema from the package.json.

Regards,
Ludwig

daogrady:
Hi Ludwig,

> Would it be possible to totally keep the schema from the package.json in git and just add it via script call right before the npm publish step in the release workflow?

Sure, that's also possible. I am always a bit anxious to release something I have not seen physically manifested and then have something different in the registry than in the repository, but I guess it's not much worse than webpacking or minifying. So I am not generally opposed to this approach.

> I made the changes

Looks great, thanks for the effort! I will bring up the automated generation of the schema in _package.json_ in the next typer sync.

Best,
Daniel

daogrady (approved):

daogrady:
Hi Ludwig,

just fyi, I now had the opportunity to ask the colleague about the camelCase-snake_case-confusion. The problem arises when users specify environment variables, which are by convention in UPPERCASE (`MYVAR=42 cds ...`). There is no clear mapping onto a `cds.env` variable `myVar` (or should it be `mYvar`? `mYVar`?, `myvAr`?). So it has become a convention to use snake_case for configurable options, which eventually arrived in cds-typer where we had already used camelCase for a while. So that's why we are now supporting both, camelCase, as well as snake_case.

Best,
Daniel

PR #345, opened 2024-10-04
feat: add static property `elements` to entity classes

stockbal:
Enables proper access to all entity elements of the underlying `LinkedDefinition`.

e.g.

```cds
entity Books : cuid {
  title : String;
  stock : Integer;
}
```

new property `elements` allows access to all defined elements of the entity and access the underlying `LinkedDefinition` of those elements.

```ts
import { Book } from "#cds-models/my/bookshop";
import cds from "@sap/cds";

if (Book.elements.ID instanceof cds.builtin.classes.UUID) console.log("ID is of type UUID");
if (Book.elements.stock instanceof cds.builtin.classes.Integer) console.log("stock is of type Integer");
```

### DIFF
--- a/lib/components/basedefs.js
+++ b/lib/components/basedefs.js
@@ -15,7 +15,7 @@ const baseDefinitions = new SourceFile('_')
 baseDefinitions.addPreamble(`
 import { entity } from '@sap/cds'
 
-export type ElementsOf<T> = {[name in keyof DeepRequired<T>]: entity["elements"][string]}
+export type ElementsOf<T> = {[name in keyof Required<T>]: entity["elements"][string]}
 
 export namespace Association {
     export type to <T> = T;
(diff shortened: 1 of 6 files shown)


lib/components/basedefs.js
```diff
@@ -13,6 +13,10 @@ const timeRegex = '`${number}${number}:${number}${number}:${number}${number}`'
 const baseDefinitions = new SourceFile('_')
 // FIXME: this should be a library someday
 baseDefinitions.addPreamble(`
+import { entity } from '@sap/cds'
+
+export type ElementsOf<T> = {[name in keyof Required<T>]: entity["elements"][string]}
```
daogrady, line 18: this should be the same as
```ts
import { type } from '@sap/cds'

export type ElementsOf<T> = {[name in keyof Required<T>]: type}
```
, or not?

stockbal: You are right. I will make the change

stockbal: You are right. I will make the change

daogrady (approved):
Looks good, thanks for this addition! :)

PR #346, opened 2024-10-04
fix: entity element with named struct types not flattened

stockbal:


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 - Schema definition for `cds.typer` options in `package.json` and `.cdsrc-*.json` files
 
+### Fixed
+- Entity elements of named structured types are flatted when using the option `--inlineDeclarations flat`
```
daogrady, line 12:
```suggestion
- Entity elements of named structured types are flattened when using the option `--inlineDeclarations flat`
```

lib/visitor.js
```diff
@@ -589,17 +593,19 @@ class Visitor {
      * @param {EntityCSN} element - CSN data belonging to the the element.
      * @param {SourceFile} file - the namespace file the surrounding entity is being printed into.
      * @param {Buffer} [buffer] - buffer to add the definition to. If no buffer is passed, the passed file's class buffer is used instead.
+     * @param {TypeResolveOptions} [resolverOptions] - custom type resolver options
      * @returns @see InlineDeclarationResolver.visitElement
      */
-    visitElement(name, element, file, buffer = file.classes) {
+    visitElement(name, element, file, buffer = file.classes, resolverOptions) {
```
daogrady, line 599: This feels a bit awry, as it introduces an optional parameter without explicit default value after ones with explicit default values (sonarlint complains about this as well).
I think this should be streamlined with [`InlineDeclarationResolver::visitElement`](https://github.com/cap-js/cds-typer/blob/main/lib/components/inline.js#L78) where we pass a destructured object instead. That was due before your PR, but with the introduction of yet another parameter, this feels even more awkward. Could you please incorporate this change?

daogrady (changes requested):
Thanks for this contribution! Looks mostly good, just one minor request wrt passing the newly introduced parameter and then we're good to go 🚀

lib/visitor.js
stockbal: Agreed. I made the adjustment

daogrady (approved):

PR #348, opened 2024-10-05
fix: creates `drafts` property only at entities that are truly draft enabled

### LINKED ISSUE #68: [IMPROVEMENT] Remove `.drafts` Property From Entities That Are Not Actually Draftable
Follow-up to: https://github.com/cap-js/cds-typer/issues/34

`.drafts` should only be present in entities that can actually be draftable. Currently, entities that aren't actually draftable might end up with a `.drafts` property.

- [ ] check with the compiler team whether there is a surefire way to determine an entity's draftability
- [ ] if possible, get rid of the current voodoo where we propagate annotations through the CSN
- [ ] make sure a static `.drafts` property is only exposed when the entity can actually be drafted during runtime

stockbal:
Uses a new approach taken from [@sap/cds-compiler](https://www.npmjs.com/package/@sap/cds-compiler)/lib/transform/draft.js#generateDraft to detect draft enabled entities.

In essence there are only 2 cases when a draft table will be generated for an entity and therefore the `.drafts` property of an entity will return a valid CSN entity reference during runtime.

1) Entity A is exposed in a service and has the annotation `@odata.draft.enabled`
2) Entity B is used as composition in an entity A, which in turn is exposed in a service and annotated with `@odata.draft.enabled`

### Further remarks

- The refactoring uses the `inferred` csn model, because the `xtended` one does not include the autoexposed entities.
- The draft state is no longer propagated top-down, as the model entities must not get the `drafts` properties. Instead, all draft enabled _service_-entities are collected in an array

Fixes #68


lib/csn.js
```diff
@@ -87,183 +91,144 @@ const getProjectionTarget = entity => isProjection(entity)
...
+            /** @type {Record<string,EntityCSN>} */
+            const draftEntities = {}
+            this.#collectDraftNodesInto(draftRoot, draftRoot.name, draftRoot, draftEntities)
 
-        this.#propagateProjections()
+            for (const draftNode of Object.values(draftEntities)) {
+                draftEnabledEntities.push(draftNode.name)
+            }
+        }
+        if (this.#compileError) throw new Error('Compilation of model failed')
```
daogrady, line 191: why has this become necessary?

lib/csn.js
```diff
@@ -87,183 +91,144 @@ const getProjectionTarget = entity => isProjection(entity)
...
 // note to self: following doc uses ＠ homoglyph instead of @, as the latter apparently has special semantics in code listings
 /**
- * We are unrolling the @odata.draft.enabled annotations into related entities manually.
- * This includes three scenarios:
+ * We collect all entities that are draft enabled.
+ * (@see `@sap/cds-compiler/lib/transform/draft/db.js#generateDraft`)
  *
- * (a) aspects via `A: B`, where `B` is draft enabled.
- * Note that when an entity extends two other entities of which one has drafts enabled and
- * one has not, then the one that is later in the list of mixins "wins":
```
daogrady, line 235: I guess you don't need to consider ancestry, as the inferred model already has the annotations propagated to the appropriate places for inheritance, is that correct?

lib/visitor.js
```diff
@@ -297,16 +299,17 @@ class Visitor {
...
-        buffer.add(`export class ${identSingular(clean)} extends ${identAspect(toLocalIdent({clean, fq}))}(${baseDefinitions.path.asIdentifier()}.Entity) {${this.#staticClassContents(clean, entity).join('\n')}}`)
+        buffer.add(`export class ${identSingular(clean)} extends ${identAspect(toLocalIdent({clean, fq}))}(${baseDefinitions.path.asIdentifier()}.Entity) {${this.#staticClassContents(fq, clean, entity).join('\n')}}`)
         this.contexts.pop()
     }
 
     /**
+     * @param {string} fq - fully qualified name of the entity
      * @param {string} clean - the clean name of the entity
-     * @param {EntityCSN} entity - the entity to generate the static contents for
+     * @param {EntityCSN} _entity - the entity to generate the static contents for
```
daogrady, line 309:
```suggestion
```

lib/visitor.js
```diff
@@ -297,16 +299,17 @@ class Visitor {
...
      */
-    #staticClassContents(clean, entity) {
-        return isDraftEnabled(entity) ? [`static drafts: typeof ${clean}`] : []
+    #staticClassContents(fq, clean, _entity) {
```
daogrady, line 311:
```suggestion
    #staticClassContents(fq, clean) {
```

lib/visitor.js
```diff
@@ -380,7 +380,7 @@ class Visitor {
             }
             // plural can not be a type alias to $singular[] but needs to be a proper class instead,
             // so it can get passed as value to CQL functions.
-            const additionalProperties = this.#staticClassContents(singular, entity)
+            const additionalProperties = this.#staticClassContents(fq, singular, entity)
```
daogrady, line 383:
```suggestion
            const additionalProperties = this.#staticClassContents(fq, singular)
```

test/util.js
```diff
@@ -183,7 +183,8 @@ async function prepareUnitTest(model, outputDirectory, parameters = {}) {
         await checkTranspilation(tsFiles)
     }
     configuration.setFrom(configurationBefore)
-    return { astw: new ASTWrapper(path.join(parameters.fileSelector(paths), 'index.ts')), paths }
+    // return undefined for astw in case type-generation failed
+    return { astw: paths?.length > 0 ? new ASTWrapper(path.join(parameters.fileSelector(paths), 'index.ts')) : undefined, paths }
```
daogrady, line 187: that was a "nice" (read: hard to love) way of finding when tests failed due to compilation errors, i.e. when the test model contained syntax errors, etc.
What are the implications of returning `undefined` here? Maybe we should take this opportunity to make the fail-state implicit and have something like

```suggestion
    if (!paths?.length) throw new Error('type-generation failed')
    return { astw: new ASTWrapper(path.join(parameters.fileSelector(paths), 'index.ts')), paths }
```
?

lib/csn.js
stockbal: Exactly.

lib/csn.js
stockbal: I wouldn't say necessary. I see it more as an early fail-safe. If one of the following commands with a _invalid_ draft model would be executed 
- `cds compile srv -4 odata`
- `cds compile srv -4 sql`
- `cds w`

you would get the same outcome. The compilation fails.

But if you think typer should only output a log error and just continue (without generating the `drafts` property), then I change the code.

test/util.js
stockbal: Yes, in retrospect, throwing an error is clearer and would also work for the negative test case in `drafts.test.js`

```js
        await expect(
            prepareUnitTest('draft/error-catalog-service.cds', locations.testOutput('bookshop_projection'), {
                typerOptions: { logLevel: 'ERROR' },
            })
        ).rejects.toThrow('type-generation failed')
```

stockbal: We could also just remove the `.catch` clause for the `cds2ts` call. This would propagate the thrown errors inside csd-typer to the caller of `prepareUnitTest`

daogrady: Now that I think about it, I believe there was a deeper reason for just having an implicit fail 🤔
I seem to remember that explicitly throwing or removing `.catch` would just halt the entire test suite on the first failing test, instead of showing all failing tests. But I don't remember for sure...

stockbal: Hm. I made a few tests where I broke some cds models in `/test/unit/files`, but only those were failing then. The rest of the test suites ran without any issues

daogrady: Well, regardless, I think your suggested changes are objectively better than having arcane tribal knowledge of why tests should fail in certain constellations. So I guess as long as the tests run as expected, this is an improvement. 👍

lib/csn.js
daogrady: I think it is fair to bail in that case. Could you please just add a comment to briefly outline when this could fail? (I.e. like the commands you listed above)

daogrady:
Admittedly, I lack the deeper user-side understanding required to evaluate if this fully covers all cases, but as the current approach is just a heuristic, and your approach eliminates some of the use of extended CSN, I am strongly in favour of rolling with it to see if gaps pop up.
Thanks for the contribution! :)

daogrady (approved):

Issue #100, opened 2023-10-26
[BUG] Drafts - types are missing draft-management fields


geert-janklaps:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

We're in the process of migrating our existing types generated with version 0.5.0 to the new typer functionality. Since this update types have become more strict which now seems to be causing issues when trying to work with drafts entities. (e.g. inserting additional draft data through code)

Each draft entity has four fields for draft management:
- DraftAdministrativeData_DraftUUID
- HasDraftEntity
- HasActiveEntity
- IsActiveEntity

Currently it seems like these fields are not covered by the typer functionality.

### Expected Behavior

Having these administrative fields available, so we can access the draft_uuid to e.g. add additional entities to a composition through code.

### Environment

```markdown
OS: Business application studio
Versions:
@cap-js/cds-typer: 0.11.1
@cap-js/db-service: 1.3.2
@cap-js/sqlite: 1.3.1
@sap/cds: 7.3.1
@sap/cds-common-content: 1.2.0
@sap/cds-compiler: 4.3.2
@sap/cds-dk: 7.3.1
@sap/cds-dk (global): 7.3.1
@sap/cds-fiori: 1.1.0
@sap/cds-foss: 4.0.2
@sap/cds-hana: 2.0.0
@sap/cds-mtxs: 1.12.1
@sap/eslint-plugin-cds: 2.6.3
Node.js: v18.14.2
```

daogrady:
Hi Geert-Jan,

thank you for reporting this issue. Draftability itself is a bit fragile in cds-typer and will require a rework somewhere down the line. As I am not very familiar with drafts myself, could you maybe provide a minimal working sample where I could observe these four missing fields? I fail to find them anywhere in the compiled CSN.

Best,
Daniel

Issue #288, opened 2024-07-31
[BUG] A generated index.ts files has syntax errors "Duplicate Identifier ..."


D027152:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

When generating files in @cds-models, one of the generated files has syntac errors. e.g.

> Cannot redeclare exported variable '_PaymentAgreementOutgoingAspect'.
> Duplicate function implementation.
> Duplicate identifier 'PaymentAgreementOutgoing'.
> Class 'PaymentAgreementOutgoing' used before its declaration.

The file is this one: https://github.tools.sap/erp4sme/crypto-for-business/blob/switch-to-cds-typer-payment-master-data/%40cds-models/sap/erp4sme/c4b/masterData/businessPartners/index.ts

### Expected Behavior

Files are generated without syntax errors

### Steps To Reproduce

repo https://github.tools.sap/erp4sme/crypto-for-business
branch switch-to-cds-typer-payment-master-data
above-mentioned files are located in https://github.tools.sap/erp4sme/crypto-for-business/tree/switch-to-cds-typer-payment-master-data/%40cds-models)

to re-generate the files run npm run cds-typer:dev

### Environment

```markdown
| c4b                    | github.tools.sap/erp4sme/crypto-for-business.git                           |
| ---------------------- | -------------------------------------------------------------------------- |
| @cap-js/audit-logging  | 0.8.0                                                                      |
| @cap-js/cds-types      | 0.2.0                                                                      |
| @cap-js/change-trackin | 1.0.6                                                                      |
| @cap-js/telemetry      | 0.2.3                                                                      |
| @sap/cds               | 7.9.4                                                                      |
| @sap/cds-common-conten | 2.0.0                                                                      |
| @sap/cds-compiler      | 4.9.4                                                                      |
| @sap/cds-dk            | 7.9.6                                                                      |
| @sap/cds-dk (global)   | 7.7.2                                                                      |
| @sap/cds-fiori         | 1.2.3                                                                      |
| @sap/cds-foss          | 5.0.0                                                                      |
| @sap/cds-mtxs          | 2.0.3                                                                      |
| @sap/eslint-plugin-cds | 3.0.4                                                                      |
| Node.js                | v18.18.0                                                                   |
| home                   | /Users/D027152/Documents/git/dch/crypto-for-business/node_modules/@sap/cds |

npx cds-typer --version: 0.23.0
```

[...]

PR #354, opened 2024-10-07
feat: enhance `.drafts` properties with draft table properties

### LINKED ISSUE #163: `Plural.drafts` could be `typeof Plural`

stockbal:
Hi @daogrady,

this is a proposal for enhancing the type behind the `.drafts` property of draft enabled classes, to include the additional fields that will be generated for draft tables.

At the same time it would also overwork `.drafts` at the plural entity, to use an enhanced version of the generated plural class of an entity, so fixes #163.

P.S.: tests do not yet exist. I will create them if you think we can go ahead with this approach

### Tasks

- [x] add test cases

### Sample

```cds
// cat-service.cds
service CatalogService {
  @odata.draft.enabled
  entity Books : cuid { }
}
```
#### Generated classes

```ts
// /_/index.ts
export type DraftEntity<T> = T & {
    IsActiveEntity?: boolean | null
    HasActiveEntity?: boolean | null
    HasDraftEntity?: boolean | null
    DraftAdministrativeData_DraftUUID?: string | null
}
export type DraftOf<T> = { new(...args: any[]): DraftEntity<T> }
export type DraftsOf<T> = typeof Array<DraftEntity<T>>

// /CatalogService/index.ts
export class Book extends _BookAspect(__.Entity) {static drafts: __.DraftOf<Book>}
export class Books extends Array<Book> {static drafts: __.DraftsOf<Book>}
```

#### Explanation about types

The type `DraftEntity<T>`  is exported so users can type their variables or parameters to a draft entity

```ts
import { Book } from "#cds-models/CatalogService";
import { DraftEntity } from "#cds-models/_";

function getBookDraft(): DraftEntity<Book> {}
```

#### Side note

Fixes auto-closed issue #100


lib/visitor.js
```diff
@@ -280,19 +282,44 @@ class Visitor {
...
+     * @param {string} fq - FQN of entity
+     * @param {Buffer} buffer - buffer to write exported drafts classes to
+     * @param {string} singular - singular name of parent entity
+     * @param {string} aspectFunction - class creator function of parent entity
+     */
+    #printDraftClasses(fq, buffer, singular, aspectFunction) {
+        if (!isDraftEnabled(fq)) return
+        const singularDraft = `${singular}Draft`
+        // add singular draft
+        buffer.addIndentedBlock(`export class ${singularDraft} extends ${aspectFunction}(__.DraftEntity) {`, () => {
```
daogrady, line 306: I am always very, very sceptical when we try to introduce identifiers in the types that are not actually in the CSN. It's already sometimes an issue with automatic inflection, and has recently [become an issue with services](https://github.com/cap-js/cds-typer/issues/288) that are now also generated into the _index.ts_, which formerly represented the service itself. (I promise you there will be someone who has an entity `Foo` _and_ another entity `FooDraft` 😬)

stockbal: I am totally with you. The suffix was merely a suggestion and I am not married to it 😊. We could make the name configurable via annotations (`@draftSingular`, `@draftPlural`), or globally via CLI option (`draft_class_suffix`) or another approach that would allow the user to fix the types if their model falls into an edge case.

If you do not like the dedicated class approach for the drafts at all, we can skip them entirely and just enhance the type behind the `.drafts` property to include the properties of a draft entity.

```ts
// /@cds-models/_/index.ts
type DraftEntity<T> = T & {
    IsActiveEntity?: boolean | null
    HasActiveEntity?: boolean | null
    HasDraftEntity?: boolean | null
    DraftAdministrativeData_DraftUUID?: string | null
}
export type DraftOf<T> = { new(...args: any[]): DraftEntity<T> }
export type DraftsOf<T> = typeof Array<DraftEntity<T>>

// /@cds-models/CatalogService/index.ts
import * as __ from './../_';

export class Book extends _BookAspect(__.Entity) {static drafts: __.DraftOf<Book>}
export class Books extends Array<Book> {static drafts: __.DraftsOf<Book>}
```

daogrady: > just enhance the type behind the .drafts property to include the properties of a draft entity.

I think this would be the most preferable approach. Introducing CSN annotations always has to be approved internally first. And I'd prefer to introduce CLI options sparingly. So enhancing the type sounds very appealing to me.

stockbal: I made the changes. One downside of omitting the dedicated draft classes was the advantage of easy type casts.

e.g.

```ts
INSERT.into(Books.drafts).entries({IsActiveEntity: true} as BookDraft);

let bookDraft = {} as  BookDraft;
bookDraft.IsActiveEntity = true
```

So the user still can do this in a rather easy way I exported the new `DraftEntity` type from the generated `/_/index.ts` file.

```ts
INSERT.into(Books.drafts).entries({IsActiveEntity: true} as DraftEntity<Book>);

let bookDraft = {} as  DraftEntity<Book>;
bookDraft.IsActiveEntity = true
```

lib/visitor.js
```diff
@@ -336,9 +336,12 @@ class Visitor {
     /**
      * @param {string} fq - fully qualified name of the entity
      * @param {string} clean - the clean name of the entity
+     * @param {boolean} [isPlural] - `true` if passed entity is plural
      */
-    #staticClassContents(fq, clean) {
-        return isDraftEnabled(fq) ? [`static drafts: typeof ${clean}`] : []
+    #staticClassContents(fq, clean, isPlural = false) {
+        if (!isDraftEnabled(fq)) return []
+        const draftsType = isPlural ? '__.DraftsOf' : '__.DraftOf'
```
daogrady, line 343: Please add respective [wrappers](https://github.com/cap-js/cds-typer/blob/main/lib/components/wrappers.js) for these two code fragments for consistency

lib/visitor.js
daogrady: That sounds reasonable and totally doable for users when they want to reuse this type. 👍

daogrady (approved):
Looks good, thanks for yet another cool contribution! 🙂

PR #355, opened 2024-10-08
Enable service names with non-alphanum characters

daogrady:
Allows "proper" naming of service containing non-alphanumeric names:

```cds
service "Foo/Bar" {}
```

would become

```ts
class __Foo_Bar {}  // not exported directly
export { __Foo_Bar as "Foo/Bar" }
```

and users would have to import like so:

```ts
import { "Foo/Bar" as FooBar } from ...
//                      ^ any alias
```

Note that this creates a discrepancy between service that don't contain special characters, as they are exported as `default` export, and ones that have names containing special characters, that are only exported in named style.

- [ ] add tests
- [ ] output warning when such a service name is used, as this [requires](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#support-for-arbitrary-module-identifiers) the use of typescript@5.6 or higher. -> only debug message

### DIFF
--- a/lib/components/identifier.js
+++ b/lib/components/identifier.js
@@ -17,7 +17,24 @@ const normalise = ident => ident && !isValidIdent.test(ident)
  */
 const last = ident => ident.split('.').at(-1) ?? ident
 
+/**
+ * Normalises the name of a service.
+ * @param {string} name - name of the service or fq thereof.
+ */
+const normalisedServiceName = name => {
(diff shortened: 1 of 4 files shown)


daogrady:
PR has become superfluous with https://github.com/cap-js/cds-typer/pull/454 but will be partially reused for entities with special characters in https://github.com/cap-js/cds-typer/pull/455

Issue #298, opened 2024-08-21
[BUG] Erroneous generated entities when overriding properties of parent


aschmidt93:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

When an entity extends another entity and changes the value of a property, the generated typescript code is erroneous. 
Only in versions > 0.22

### Steps To Reproduce

```cds
type OrderState         : String enum {
        Ready    = 'READY';
    };

    type FinishedOrderState : String enum {
        Done     = 'DONE';
        Canceled = 'CANCELED';
    }

    entity Orders {
        key ID          : UUID;
            state       : OrderState;
            predecessor : Association to Orders;
    };

    entity FinishedOrders : Orders {
        state       : FinishedOrderState;
        predecessor : Association to FinishedOrders;
    }
```

Generates 
![image](https://github.com/user-attachments/assets/1d227e45-f588-4728-b101-f56f348dace2)

### Environment

```markdown
- **cds-typer**: >0.22
- **cds**: ^7
```

daogrady:
Hi Andreas,

thanks for reporting this issue!
To be honest, I wouldn't know how a matching type definition would look like and I feel like this should not be allowed in CDS. @johannes-vogel could you please weigh in and let me know if extending and entity and thereby changing the type of a property is desired behaviour?

Minimal example:

```cds
entity A {
  foo: Integer
};

entity B: A {
  foo: String
}
```

Best,
Daniel

hakimio:
Looking just at TypeScript types, the overrides are incompatible. All those cases should result in TS errors.
- `OrderState` is incompatible with `FinishedOrderState`.
- Same goes for `string` and `number` example.

Here is how corrected TS types could look like:
```ts
enum OrderState {
    Ready    = 'READY',
    Done     = 'DONE',
    Canceled = 'CANCELED'
};

interface Order {
        ID          : UUID;
        state       : OrderState;
        predecessor : Order;
};

interface FinishedOrder: Order {
    state       : OrderState.Done | OrderState.Canceled;
    predecessor : FinishedOrder;
}
```

daogrady:
Hi Andreas,

we looked into the issue and I am afraid this is a capability of CDS that is not really expressible in TypeScript. 
I don't see a way to accommodate this type of model in general, as it allows for true type-changes (i.e. `string` becomes `number`, and so on), not just for narrowing, as in the solution Tomas suggested.

[...]

PR #356, opened 2024-10-12
feat/refactor: use one csn only + other improvements

### LINKED ISSUE #77: [FEATURE] Missing types for text tables/localized properties
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

When an entity property is defined with "localized", there is another table created that contains "locale" as additional key, the original entity key fields + all localized properties. It would be beneficial if a new class was generated that represents entries in this table as well as a new reference to this text table that can be used in CQL commands. 

For inspiration, the project [cds2types](https://github.com/mrbandler/cds2types) does generate classes for the text entities. 

### Expected Behavior

A new classes are generated for the text tables.

### Steps To Reproduce

_No response_

### Environment

```markdown
- **OS**:
- **Node**:
- **npm**:
- **cds-typer**: 
- **cds**:
```

### Repository Containing a Minimal Reproducible Example

_No response_

### Anything else?

_No response_

### LINKED ISSUE #116: [BUG] String enum in composition of many potentially not supported
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

Given cds such as:

```cds
entity Something : cuid {
  taggedElements : Composition of many {
                                  hierarchyLevel : String enum {
                                    oneTaggedElementHierarchyLevel
                                  };
                              key elementGuid           : String;
                            };
```

The following type is generated:

```ts
export function _SomethingAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
  return class Something extends Base {
        taggedElements?: __.Composition.of.many<Array< {
  hierarchyLevel?: taggedElements_hierarchyLevel | null,
  elementGuid?: string,
}>>;
  };
}
```

Yet `taggedElements_hierarchyLevel` is not defined.

### Expected Behavior

I would've expected `taggedElements_hierarchyLevel` to exist.

### Steps To Reproduce

_No response_

### Environment

```markdown
- **OS**: Mac OS
- **Node**: v20.10.0
- **npm**: 10.2.3
- **cds-typer**: 0.12.0
- **cds**: 7.4.1
```

### Repository Containing a Minimal Reproducible Example

_No response_

### Anything else?

_No response_

### LINKED ISSUE #128: [FEATURE] Consolidate Consumed CSN Flavours
## Two Flavours

cds-typer nowadays uses two flavours of CSN to operate: xtended and inferred flavour.
The inferred flavour is generally superior in our case, as it contains fields that are computed later down the pipeline, like properties from CQN selections, annotations, and so on. These are all fields we are expecting in our generated type files.

The xtended flavour on the other hand does not propagate fields from inherited entities to child entities. Consider the following CDS file:

```cds
entity A {
    id : Integer;
}

entity B: A {
    name: String;
}
```

Loading it with the inferred flavour will generate this CSN snippet:

```csn
    A: {
      kind: 'entity',
      elements: { id: { type: 'cds.Integer' } }
    },
    B: {
      kind: 'entity',
      includes: [ 'A' ],
      elements: {
        id: { type: 'cds.Integer' },
        name: { type: 'cds.String' }
      }
    }
  }
```

which would (more or less) result in the following classes:

```ts
class A {
  id: number;
}

class B extends A {
  id: number;  // <- duplicate!
  name: string;
}
```

This duplicate is not present when using the xtended flavour instead.

## Ramifications

As the compilation step is not discrete and/pure, we need to load the entire model twice. This can be a considerable hit on the execution time.
Having both flavours can also be rather confusing. We constantly need to juggle them and remember which flavour to use in which use case.

### Suggested Solution

We should get rid of one of the two flavours. As inferred CSN is more complete, the xtended flavour should probably be dropped. Instead of rectifying the xtended flavour with unrolled annotations, we should aim to deduplicate properties in the inferred flavour instead.

### Alternatives

Wait for universal CSN or modify xtended CSN even more to be closer to inferred CSN.

### Additional Context

_No response_

stockbal:
Hi @daogrady,

proposal to get rid of one of the CSN flavors. I chose to remove the `xtended` flavor. All in all the required fixes/changes were not many and had some nice side effects.

- working dedicated classes for inline compositions 
- working dedicated classes for `localized` elements

Here is short sample to show the changes to the generated classes (only the important parts are included)

### Sample cds schema
```cds
namespace bookshop;

entity Books : cuid {
  title      : localized String;
  publishers : Composition of many {
                 key ID      : UUID;
                     name    : String;
                     type    : String enum {
                       self;
                       independent;
                     };
                     offices : Composition of many {
                                 key ID      : UUID;
                                     city    : String;
                                     zipCode : String;
                               }
               }
}
```

### Generated `index.ts` for namespace `bookshop`

```ts
// enum
const publisher_type = {
  self: "self",
  independent: "independent",
} as const;
type publisher_type = "self" | "independent"

export function _BookAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
  return class Book extends _._cuidAspect(Base) {
    declare title?: string | null;
    declare publishers?: __.Composition.of.many<Books.publishers>;
    // properties to access composition and association to generated table for 'localized' elements
    // #################################
    declare texts?: __.Composition.of.many<Books.texts>;
    declare localized?: __.Association.to<Books.text> | null;
    // #################################
  };
}
export class Book extends _BookAspect(__.Entity) {}
export class Books extends Array<Book> {$count?: number}

export namespace Books {
  // classes for inline composition inside namespace of parent entity
  export function _publisherAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
    return class publisher extends Base {
      declare up_?: __.Key<__.Association.to<Book>>;
      declare up__ID?: __.Key<string>;
      declare ID?: __.Key<string>;
      declare name?: string | null;
      declare type?: publisher_type | null;
      declare offices?: __.Composition.of.many<Books.publishers.offices>;
      
      static type = publisher_type
    };
  }
  export class publisher extends _publisherAspect(__.Entity) {}
  export class publishers extends Array<publisher> {$count?: number}
  
  // generated text entity inside namespace of parent entity
  export function _textAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
    return class text extends _sap_common._TextsAspectAspect(Base) {
      declare ID?: __.Key<string>;
      declare title?: string | null;
    };
  }
  export class text extends _textAspect(__.Entity) {}
  export class texts extends Array<text> {$count?: number}
  
}

export namespace Books.publishers {
  // classes for nested inline composition 'offices'
  export function _officeAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
    return class office extends Base {
      declare up_?: __.Key<__.Association.to<Books.publisher>>;
      declare up__ID?: __.Key<string>;
      declare up__up__ID?: __.Key<string>;
      declare ID?: __.Key<string>;
      declare city?: string | null;
      declare zipCode?: string | null;
    };
  }
  export class office extends _officeAspect(__.Entity) {}
  export class offices extends Array<office> {$count?: number}
}
```
### Generated `index.js` for namespace `bookshop`

```js
// Books
module.exports.Book = { is_singular: true, __proto__: csn.Books }
module.exports.Books = csn.Books
// Books.publishers
module.exports.Books.publisher = { is_singular: true, __proto__: csn.Books.publishers }
module.exports.Books.publishers = csn.Books.publishers
// Books.texts
module.exports.Books.text = { is_singular: true, __proto__: csn.Books.texts }
module.exports.Books.texts = csn.Books.texts
// Books.publishers.offices
module.exports.Books.publishers.office = { is_singular: true, __proto__: csn.Books.publishers.offices }
module.exports.Books.publishers.offices = csn.Books.publishers.offices
// events
// actions
// enums
module.exports.publisher.type ??= { self: "self", independent: "independent" }
```

This branch is again based on the fix for the draftable state (see #348)

Fixes #116 
Closes #77
Closes #128 

Let me know if you want to go forward with this branch/approach.

Regards,
Ludwig


daogrady:
Hi Ludwig,

very impressive work, thanks for putting in the effort! As this is quite the intrusive change, I'd like to give this a more thorough look. I will be out of office for two weeks, starting Friday, so I won't be able to look into this (or other PRs) until mid November. Just so you know this is not being shelved, just put on hold until I have the time to give it a proper review. 🙂

Best,
Daniel

lib/visitor.js
```diff
@@ -333,6 +288,22 @@ class Visitor {
         this.contexts.pop()
     }
 
+    /**
+     * @param {EntityCSN} entity -
+     * @returns {string[]}
+     */
+    #getInheritedElements(entity) {
```
daogrady, line 295: This smells a bit like a bottle neck (both, the finding of these elements and using an array over, say, `Set`) and something that could be cached as part of `EntityInfo`. But we can probably benchmark it at some later point.

daogrady:
One particular problem we formerly overcame with our two-flavour-approach was dealing with the provenance of annotations. This is, as far as I am aware, generally an unsolved problem when you don't have access to the xtended flavour. Consider the following model:

```cds
@singular: 'A'
entity A {}

entity B: A {}
```

Using only the inferred flavour, both `A` and `B` would have the `@singular: 'A'` annotation. The generated code would therefore contain a duplicate class `A`:

```ts
// This is an automatically generated file. Please do not change its contents manually!
import * as __ from './_';

export function _AAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
  return class A extends Base {
    static readonly kind: "entity" | "type" | "aspect" = 'entity';
    declare static readonly keys: __.KeysOf<A>;
    declare static readonly elements: __.ElementsOf<A>;
    static readonly actions: Record<never, never>;
  };
}
export class A extends _AAspect(__.Entity) {}
Object.defineProperty(A, 'name', { value: 'A' })
Object.defineProperty(A, 'is_singular', { value: true })
export class A_ extends Array<A> {$count?: number}
Object.defineProperty(A_, 'name', { value: 'A' })

export function _AAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
  return class A extends _AAspect(Base) {
    declare x?: number | null
    static override readonly kind: "entity" | "type" | "aspect" = 'entity';
    declare static readonly keys: __.KeysOf<A>;
    declare static readonly elements: __.ElementsOf<A>;
    static readonly actions: typeof A.actions & Record<never, never>;
  };
}
// v ❌ v
export class A extends _AAspect(__.Entity) {}
Object.defineProperty(A, 'name', { value: 'B' })
Object.defineProperty(A, 'is_singular', { value: true })
export class B extends Array<A> {$count?: number}
Object.defineProperty(B, 'name', { value: 'B' })
```

did you consider this aspect already or do you see a way to handle this with only the inferred flavour available?

stockbal:
To be honest I didn't think of that. Considerung your sample I personally would say that such entity inheritence scenarios are not very realistic. Until now, I only encountered the use of aspects to enrich other entities with a set of annotations or fields (e.g. cuid, managed, CodeList to name a few). In that case we would be safe as one would not add `@singular` or `@plural` to an aspect.

That being said, the `inferred` model contains the annotations according to the rules of [annotation propagation](https://cap.cloud.sap/docs/cds/cdl#annotation-propagation). In the past I personally had some issues because we had to annotate our entities in the database model and again in the service because of the missing propagation of `@singular/@plural`.

So there a times a propagation of `@singular/@plural` is welcome and times where it is not.

e.g.

```cds
// schema.cds
namespace db;

// -> MyBooks, MyBook
@singular: 'MyBook'
@plural: 'MyBooks'
entity Books {}

// -> Pubs, Pub
@singular: 'Pub'
@plural: 'Pubs'
entity Publishers {}

// -> BestsellingBooks, BestsellingBook
@singular: null // stop propagation and keep original name of entity
@plural: null // stop propagation and keep original name of entity
entity BestsellingBooks as select * from Books {}
```

```cds
// service.cds
using {db} from '../db/schema';

service MyService {
   // -> Books, Book
   @singular: null // stop propagation and keep original name of entity
   @plural: null // stop propagation and keep original name of entity
   entity Books as projection on db.Books;
   
   // -> BestsellingBooks, BestsellingBook (annotation propagation was stopped)
   entity BestsellingBooks as projection on db.BestsellingBooks;
   
   // -> Pubs, Pubs (via propagation)
   entity Publishers as projection on db.Publishers;
}
```

At the end, it is of course a breaking change and users would need to adjust their models to correct their generated types, but cds-typer would follow the documented rules of annotations are propagated. At times it may force users to add more annotations to their models as before.

So **1)** keep the xtended model for knowing the true origin of `@singular/@plural` and spare the users some thinking/writing or **2)** add some additional documentation so it is clear how the annotations would affect the generated types. I don't think there is a _good_ third option.

daogrady:
> In that case we would be safe as one would not add @singular or @plural to an aspect.

🤔 why is that? Aspects will be generated as classes by cds-typer, just as entities are. They therefore suffer from the same shortcomings our current inflection algorithm has, and thus can be annotated with custom inflection.

> In the past I personally had some issues because we had to annotate our entities in the database model and again in the service because of the missing propagation of @singular/@plural.

not sure I grasp that either. Are you talking about having to repeat the naming for service-entities, which bears the risk of being inconsistent during renamings?

As for your proposed solutions: I agree, as I don't see any third solution either. I will bring this up in the next cds-typer sync and see if there is any preferred way forward.

stockbal:
> 🤔 why is that? Aspects will be generated as classes by cds-typer, just as entities are. They therefore suffer from the same shortcomings our current inflection algorithm has, and thus can be annotated with custom inflection.

The array class type is not generated for those, only for entities. Therefore I am not even sure why inflection is required for aspects at all. For example an aspect `Status` will result in a class `Statu` right now. Which would make sense for an `entity` type because otherwise the array class would have the same name. 

e.g.:

```cds
// model.cds
aspect Status {}

entity MyStatus : Status { }
```
```ts
// index.ts
export class Statu extends _StatuAspect(__.Entity) {}

export class MyStatu extends _MyStatuAspect(__.Entity) {}
export class MyStatus extends Array<MyStatu> {$count?: number}
```



> not sure I grasp that either. Are you talking about having to repeat the naming for service-entities, which bears the risk of being inconsistent during renamings?

Exactly. I added the annotations on db level because of unsatisfactory inflected names and then had to repeat the annotations again on service level. Consistent renaming in both db and service model was actually not an issue, at least not yet. We just expected the annotations to be inherited to services level, because of the mentioned propagation

daogrady:
Hi Ludwig,

we discussed your proposal internally and we would very much like to have this change merged. Especially the simplified CSN handling, inline compositions, and localization are a great benefit in themselves.
I think we should for now keep the `xtended` flavour for looking up the location of the inflection annotations. At the same time, we acknowledge the inconvenience of "double book keeping" these annotations cause. We hope to progressively phase out the `xtended` flavour in future development.

Is that an acceptable way forward for you?

Best,
Daniel

stockbal:
Hi Daniel,

that's really great news 😊.
One idea about keeping the `xtended` csn for the inflections. I could do a very early propagation of the `singular`/`plural` annotations, i.e. remove them from the new primary model - the inferred one - and transfer them from the xtended model. After that, the xtended model instance could be discarded. I could do this inside the constructor of the Visitor class.
What do you think about that?

Regards,
Ludwig

daogrady:
Hi Ludwig,

that sounds like another step to driving back `xtendend` to just the initialisation and streamlines the use of one CSN flavour during the remaining generation process. So sounds good to me! 🙂

Best,
Daniel

daogrady:
> In that case we would be safe as one would not add @singular or @plural to an aspect.

I just realised there seems to the [hypothetical szenario](https://github.com/cap-js/cds-typer/issues/410) of adding `@singular`/ `@plural` to aspects after all. Assuming the generated singular or plural form of an aspect collides with the name of an existing entity, the user would have to add these annotations.
I believe the approach you are currently following is still robust against [generating the inflected forms](https://github.com/cap-js/cds-typer/pull/421) to cds aspects, correct?

stockbal:
I think so, as I discard the propagated annotations from every definition in the inferred model, not just entities. I will keep the _Composition of aspects_ scenario in mind.

stockbal:
Hi Daniel,

I am currently writing the tests for this PR and stumbled upon the following situation.

After this PR is merged, inline compositions will be generated inside an `export namespace {}` block.  
Let's consider the following model:

```cds
@singular: 'Bestseller'
@plural: 'Bestsellers'
entity Books {
  publishers : Composition of many {}
  ...
}
```

Right now this will result in the following generated types

```ts
export class Bestseller {}
export class Bestsellers extends Array<Bestseller> {}

// ❌ not aligned with the new name of Books
export namespace Books {
  export class publisher extends {}
  export class publishers extends Array<publisher> {}
}
```

I guess I need to somehow sync the namespace to the new name. Otherwise it could be confusing as why one needs to import `Bestsellers` to select from books, but import `Books` if one would want to select from e.g. `publishers` (i.e. `SELECT.from(Books.publishers)`).  
What do you think?

daogrady:
Hi Ludwig,

valid point. My gut feeling would be to leave the namespace at the original entity name. The inflection annotations are meant to resolve naming conflicts and give the user an escape hatch when the automatic singular-derivation-algorithm fails [^1]. For most cases, annotating either `@singular` or `@plural` should be enough to resolve the conflict. The example you showed is already somewhat unorthodox in that regard, as it is basically does a full renaming of both, the singular and plural form.
Does that make sense?

Best,
Daniel

[^1]: which, frankly, it does more often than not.

stockbal:
Yeah, that makes sense. I know my example was a bit far fetched, but if you only do a small rename (e.g. the original entity name is `List` and you wanna fix the plural to `Lists`, then the namespace is also `List` and the array class is `Lists`). 

Anyhow I just wanted to mention it, even it the situation should probably/hopefully not occur too often. So if you say it's fine, then I am good as well. 😊

Regards,
Ludwig

daogrady:
Hi Ludwig,

I absolutely see your concerns and yes, in the worst case, users would end up with up to three names for one entity in their model (singular, plural, and the namespace to top it off). But to me, the path forward you proposed makes sense. Still, I should be able to get a hold of some higher-ups today, so I can also try to get their input on the matter.

Best,
Daniel

lib/visitor.js
stockbal: Moved into entity repository

daogrady:
Hi Ludwig,

quick update: we discussed this internally and agree with your proposed solution.

Best,
Daniel

stockbal:
Hi Daniel,

ok so. And with my proposed solution you mean synching the namespace, or leaving it as is? 😊.

Regards,
Ludwig

daogrady:
Hi Ludwig,

sorry, I meant the post I had already agreed to before, but now got confirmation of the PO as well:
 
> I absolutely see your concerns and yes, in the worst case, users would end up with up to three names for one entity in their model (singular, plural, and the namespace to top it off). But to me, the path forward you proposed makes sense. Still, I should be able to get a hold of some higher-ups today, so I can also try to get their input on the matter.
 
Best,
Daniel

stockbal:
Hi Daniel,

sorry, I am still at a loss what's to be done now from my end 😅. 

P.S.: Maybe I am little slow today as I was sick the last few days.

Regards,
Ludwig

daogrady:
Hi Ludwig,

sorry for the confusion again. 😃
Please leave the namespace for scoped entities at the original name as it is found in the CSN and do not apply any `@singular`/ `@plural` rename to it.

Best,
Daniel

stockbal:
Then this PR is ready for review now

daogrady:
Hi Ludwig,

thanks for finishing this feature! I'm afraid I won't be able to give this a proper review before the holidays, as other tasks unfortunately currently take precedence, but if I don't get to it today or tomorrow, this will be one of the first things on my plate in the new year.

Best,
Daniel

lib/resolution/entity.js
```diff
@@ -61,6 +61,22 @@ class EntityInfo {
     /** @type {import('../typedefs').resolver.EntityCSN | undefined} */
     #csn
 
+    /** @type {Set<string> | null} */
+    #inheritedElements = null
+
+    /** @returns set of inherited elements (e.g. ID of aspect cuid) */
+    get inheritedElements() {
```
daogrady, line 68: Have you considered property renaming in this context? I.e., does your code exhibit well-defined behaviour for when a property is present in a parent, but also in the entity itself (but possibly with a changed type)?

```cds
entity A {
  name:String
}

entity B: A {
  name:Integer
}
```

As of today, this is a [known restriction](https://github.com/cap-js/cds-typer#changing-types). Even if you do not explicitly solve the issue, this could be an opportunity to error our if we find redefined properties.

lib/resolution/resolver.js
```diff
@@ -240,6 +242,8 @@ class Resolver {
         } else {
             // TODO: make sure the resolution still works. Currently, we only cut off the namespace!
             plural = util.getPluralAnnotation(typeInfo.csn) ?? typeInfo.plainName
+            // remove leading entity name
+            if (plural.indexOf('.') !== -1) plural = last(plural)
```
daogrady, line 246: 🌫

```suggestion
            if (plural.includes('.')) plural = last(plural)
```

lib/resolution/resolver.js
```diff
@@ -311,18 +315,6 @@ class Resolver {
                 } else {
                     let { singular, plural } = targetTypeInfo.typeInfo.inflection
 
-                    // FIXME: super hack!!
-                    // Inflection currently does not retain the scope of the entity.
```
daogrady, line 315: yay 4 fix! 🙂

lib/resolution/entity.js
```diff
@@ -61,6 +61,22 @@ class EntityInfo {
     /** @type {import('../typedefs').resolver.EntityCSN | undefined} */
     #csn
 
+    /** @type {Set<string> | null} */
+    #inheritedElements = null
```
daogrady, line 65: ⏳
```js
    /** @type {Set<string> | undefined} */
    #inheritedElements
```

probably a bit pedantic, but I use nullish values with the following distinction:

`undefined`: this has not been set yet and is therefore nothing
`null`: this has been intentionally set to nothing

So while `null` is a good default for `parent`, where we assume each entity is a "root" and therefore consciously has set its `parent` to nothing, `undefined` would fit better here, as `inheritedElements` eventually will be "something" when the getter is first called.

lib/visitor.js
```diff
@@ -74,41 +73,8 @@ class Visitor {
...
-                    // books: Association to many Author.books ...
-                    // as this would result in a type definition that
-                    // name-clashes with the actual declaration of Author
-                    this.visitEntity(name, target)
-                }
-            } else {
-                LOG.error(`Expecting an autoexposed projection within a service. Skipping ${name}`)
-            }
+        for (const [name, entity] of Object.entries(this.csn.definitions)) {
+            this.visitEntity(name, entity)
```
daogrady, line 77: yay for brevity! 🙂

daogrady (changes requested):
Thank you for this contribution!
Looks very good and even if it didn't bring support for locale classes, having weeded out several huge "FIXME this is a hack!" blocks makes this a valuable change.

I have added a few remarks and mostly annotated them with [severity markers](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/). Especially the syntax ones are largely _dust_ to me.

daogrady (changes requested):
Thank you for this contribution!
Looks very good and even if it didn't bring support for locale classes, having weeded out several huge "FIXME this is a hack!" blocks makes this a valuable change.

I have added a few remarks and mostly annotated them with [severity markers](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/). Especially the syntax ones are largely _dust_ to me, so feel free to just mark them as resolved without change if you feel strongly about your syntax choices.

lib/csn.js
```diff
@@ -292,6 +292,30 @@ function propagateForeignKeys(csn) {
...
+/**
+ * Clears "correct" singular/plural annotations from inferred model
+ * copies the ones from the xtended model.
+ *
+ * This is done to prevent potential duplicate class names because of annotation propagation.
+ * @param {{inferred: CSN, xtended: CSN}} csn - CSN models
+ */
+function propagateInflectionAnnotations(csn) {
+    const singularAnno = '@singular'
+    const pluralAnno = '@plural'
```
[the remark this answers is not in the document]
stockbal: I saw them, unfortunately I cannot use them without making additional adjustments, otherwise we will get complaints from TS.

Using it e.g. like so

```js
const singularAnno = annotations.singular[0];
const pluralAnno = annotations.plural[0];
```

results in the following issues in the `delete def[..]` statements

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'EntityCSN'.
  No index signature with a parameter of type 'string' was found on type 'EntityCSN'.ts(7053)
```

To get rid of the issues I see a few options

1. Inline JSDoc to type the anno values to existing keys of the definition
```js
const singularAnno = /** @type {'@singular'} */(annotations.singular[0])
const pluralAnno = /** @type {'@plural'} */(annotations.plural[0])
```
2. Adjusting the existing annotation constants so TS does not see them as `string[]`
```js
const annotations = {
    singular: [/** @type {const} */('@singular')],
    plural: [/** @type {const} */('@plural')],
}
```
3. Enhancing `EntityCSN` type in `typedefs.d.ts`
```ts
export type EntityCSN = Record<string, any> & {
  ...
}
```
4. Use `// ts-expect-error` for the `delete` statements

BTW, why are the annotation constants defined as arrays? Were there plans to support different versions?

lib/file.js
```diff
@@ -310,16 +313,32 @@ class SourceFile extends File {
...
+        // REVISIT: find a better way to do this???
+        const printEnumToBuffer = (/** @type {Buffer} */buffer) => printEnum(buffer, propertyToInlineEnumName(entityCleanName, propertyName), kvs, {export: false}, doc)
+
+        if (buffer?.namespace) {
+            const tempBuffer = new Buffer()
+            // we want to put the enums on class level
+            tempBuffer.indent()
+            printEnumToBuffer(tempBuffer)
+
+            // we want to write the enums at the beginning of the namespace
```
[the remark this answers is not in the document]
stockbal: You mean, why to write them at the beginning? As the other enums are also written at start of the `index.ts` file, I thought it would be constistent to also do it for enums belonging to a namespace

lib/csn.js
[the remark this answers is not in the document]
daogrady: > BTW, why are the annotation constants defined as arrays? Were there plans to support different versions?

yes, we considered supporting the singular/ plural annotations provided by Fiori, but set that idea back, as they can contain complex, localised strings that would require lookup in the i18n files.
I think your second solution using `as const` is the most suited.
Doing a superficial check, it looks like you can even do:

```js
const annotations = /** @type {const} */({
    singular: ['@singular'],
    plural: ['@plural'],
})
```

lib/file.js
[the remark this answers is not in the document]
daogrady: I see! Now that you mention it, I think enums are printed as `const` and therefore must be available in case entities refer to them in their properties. So yes, makes sense. 🙃

lib/resolution/entity.js
stockbal: No I haven't considered them - why would anyone do that 😶‍🌫️. No proper solution comes to mind right away, I would have to think about it some more. 
If I understand you correctly I could implement the following approach though, to error out and stop the type generation in case of a type mismatch.

```js
// visitor.js - #aspectify
  for (let [ename, element] of Object.entries(entity.elements ?? [])) {
    const inheritedElem = inheritedElements?.get(ename)
    if (inheritedElem) {
        const type = stringifyType(element.type)
        const inheritedElemType = stringifyType(inheritedElem.type)
        if (stringifiedType === inheritedElemType) {
            continue
        } else {
            throw new Error(`Type '${type}' of element '${entity.name}.${ename}' does not match the type '${inheritedElemType}' from the parent`)
        }
    }
  }
```

daogrady: I don't know what the architectural reasons would be, but it [has happened in the past](https://github.com/cap-js/cds-typer/issues/298). 🙂

daogrady: P.S.: I asked this mainly out of curiosity. As it is beyond the scope of your PR, this can totally be done at a later point in time and by someone else.

stockbal: Then I would actually table this for later if this is ok for you

daogrady: Absolutely! 🙂

daogrady:
@stockbal seeing that all immediate points have been addressed, do you consider this PR done and ready for merge, or do you need to make any more changes?

stockbal:
No, unless you still found something that I should change, the MR is done from my point of view.

daogrady:
Great! Then thanks once more and let's merge 🚀

daogrady (approved):

Issue #273, opened 2024-07-08
[BUG][REGRESSION] `TS4114` error introduced in `v0.23.0`


hakimio:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

Transpiling code produced by `v0.23.0` of `cds-typer` produces error:
```
X [ERROR] TS4114: This member must have an 'override' modifier because it overrides a member in the base class '_cuidAspect<{ new (...args: any[]): _managedAspect<{ new (...args: any[]): _oldCrmIdAspect<TBase>.(Anonymous class); prototype: _oldCrmIdAspect<any>.(Anonymous class); readonly actions: Record<...>; } & TBase>.(Anonymous class); prototype: _managedAspect<...>.(Anonymous class); readonly actions: Record<...>; } & { ....'. 

    ../models/CrmService/index.ts:9:8:
      9 │         ID?: string;
        ╵         ~~
```

### Expected Behavior

No errors.

### Steps To Reproduce

1. Add entity with some aspects:
```cds
entity Foo : cuid, managed {}
```
2. Generate the types
3. Try to transpile the TS code

### Environment

```markdown
- **OS**: Windows 10
- **Node**: 20.13.1
- **npm**: 10.8.1
- **cds-typer**: 0.23.0
- **cds**: 7.9.3
```

### Anything else?

[noImplicitOverride](https://www.typescriptlang.org/tsconfig/#noImplicitOverride) `tsconfig` config is set to `true`.

daogrady:
Hi Tomas,

thanks for reporting this problem! As you have correctly deduced, this problem is directly related to the `noImplicitOverride` configuration. One layover solution would be to set `noImplicitOverride` to `false`.
I fully understand (and support) that you want to make your config as strict as possible.
I'd first have to evaluate whether just slapping `override` (and `declare`) on every property has any implications. It is currently a non-trivial issue to determine whether a property already exists in a parent aspect/ entity.

So for now, I recommend removing the `noImplicitOverride` option is possible. Resolving this issue needs a bit of digging.

Best,
Daniel

hakimio:
Ok, for now I can disable `noImplicitOverride` but in the future the issue should definitely be addressed.

Anyway, another reason why I don't like the new `v0.23.0` release is that it messes up the type tooltips in the IDE.

Here is how it looked like before in `v0.22.0`:

[...]

PR #367, opened 2024-10-16
Safely override `.kind` and `Buffer` improvements

daogrady:
Only add `override` modifier to `kind` property iff any parent class actually has such a property to override.
This change is implicitly tested by having added `strict: true` to the transpilation tests in `tscheck.js`.
Fixes https://github.com/cap-js/cds-typer/issues/273

Generally, the creation of class members has been streamlined to avoid typos and incorrect order/ combination of modifiers via `lib/components.class::createMember` instead of concatenating strings manually.

Also streamlined use of `Buffer.add` to accept scalar strings, array of strings, and functions returning strings, to enable `buffer.add(xs)` over `xs.forEach(x => buffer.add(x))`.
Also added `buffer.blankLine()` to replace the somewhat obscure `buffer.add('')`

### DIFF
--- a/lib/resolution/entity.js
+++ b/lib/resolution/entity.js
@@ -61,6 +61,7 @@ class EntityInfo {
     /** @type {import('../typedefs').resolver.EntityCSN | undefined} */
     #csn
 
+    /** @returns the **inferred** csn for this entity. */
     get csn () {
         return this.#csn ??= this.#resolver.csn.definitions[this.fullyQualifiedName]
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Fixed
 - Entity elements of named structured types are flattened when using the option `--inlineDeclarations flat`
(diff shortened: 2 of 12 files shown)


lib/resolution/entity.js
```diff
@@ -61,6 +61,7 @@ class EntityInfo {
     /** @type {import('../typedefs').resolver.EntityCSN | undefined} */
     #csn
 
+    /** @returns the **inferred** csn for this entity. */
```
chgeo, line 64: Do we have a type for an (inferred) CSN?  
If not, it should ideally come from `cds-types`. (Should do this in a separate PR though).

chgeo (approved):

lib/resolution/entity.js
daogrady: We do not. The two packages were formerly strictly separated and CSN was only outlined. This could be added now that `cds-types` has become a (peer) dependency.

daogrady: I jotted down an issue https://github.com/cap-js/cds-typer/issues/368

chgeo: Yes.  `cds-types` should really become the goto-package for CDS types, not only in apps, but also in our tools.  We have more use cases in vscode-cds.

Issue #382, opened 2024-11-05
[BUG] cds types not imported for events


Janhouse:
Seems like #40 forgot to add the functionality that imports cds base types when using events.

`error TS2503: Cannot find namespace '__'.`

```ts
// This is an automatically generated file. Please do not change its contents manually!
import cds from '@sap/cds'

export class EventService extends cds.Service {
}
export default EventService

// event
export declare class Created {
  declare DistributionChannel: string | null
  declare EventRaisedDateTime: __.CdsTimestamp | null
  declare OrganizationDivision: string | null
  declare Product: string | null
  declare SalesOrder: string | null
  declare SalesOrderItem: string | null
  declare SalesOrderItemCategory: string | null
  declare SalesOrderType: string | null
  declare SalesOrganization: string | null
  declare SoldToParty: string | null
}
```

`__.CdsTimestamp` is obviously not imported and throws the error.

Janhouse:
Same problem with `action`

PR #383, opened 2024-11-05
Fix event import

GoldSucc:
Fixes https://github.com/cap-js/cds-typer/issues/382 (event import error).

### DIFF
--- a/lib/visitor.js
+++ b/lib/visitor.js
@@ -486,6 +486,7 @@ class Visitor {
         // TODO find a better way to detect ABAP RFC actions
         const isRFC = Object.values(operation.params ?? {}).some(p => Object.keys(p).some(k => k.startsWith('@RFC')))
         file.addOperation(last(fq), params, returns, kind, docify(operation.doc), {named: true, positional: !isRFC})
+        file.addImport(baseDefinitions.path)
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Janhouse:
Tested, seems to fix the issue.

chgeo:
Can you provide more context on the issue?

Janhouse:
@chgeo open the mentioned issue #382

GoldSucc:
@chgeo when cds definition contains only events/operations it generates type without __ import which is incorrect ts file. tsc produces type errors indeed because no import is present. to get this error, event/operation should use type like Timestamp (something that comes from __. like __.CdsTimestamp).

chgeo:
> @chgeo when cds definition (.cds file) contains only events/operations it generates type without __ import which is incorrect ts file. tsc produces type errors indeed because no import is present. to get this error, event/operation should use type like Timestamp (something that comes from __. like __.CdsTimestamp).

I see.  So I will add a test for that before merging.
Alternatively, you can give it a try.  See our [testing guide](https://github.com/cap-js/cds-typer/wiki/Testing) for this. Lately, we added [compilation tests](https://github.com/cap-js/cds-typer/wiki/Testing#writing-consumer-code) which are easier to write than the AST ones.

GoldSucc:
@chgeo I have some experience with code generation and syntax trees so I decided to implement tests in AST. Committed changes with tests.

Issue #391, opened 2024-11-12
[BUG] [REGRESSION] Regressions in `v0.28`: missing `declare` modifier for `.actions`


hakimio:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

Issues introduced in the latest version of the library:
- Issue with `actions` override.
Generated code:
```ts
export function _LanguageAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
  return class Language extends _CodeListAspect(Base) {
    /** Type for a language code */
    declare code?: __.Key<Locale>
    static override readonly kind: "entity" | "type" | "aspect" = 'entity';
    declare static readonly keys: __.KeysOf<Language>;
    declare static readonly elements: __.ElementsOf<Language>;
    static readonly actions: typeof CodeList.actions & Record<never, never>;
  };
}
export function _CodeListAspect<TBase extends new (...args: any[]) => object>(Base: TBase) {
  return class CodeList extends Base {
    declare name?: string | null
    declare descr?: string | null
    static readonly kind: "entity" | "type" | "aspect" = 'aspect';
    declare static readonly keys: __.KeysOf<CodeList>;
    declare static readonly elements: __.ElementsOf<CodeList>;
    static readonly actions: Record<never, never>;
  };
}
```
Error: 
```
X [ERROR] TS4114: This member must have an 'override' modifier because it overrides a member in the base class '_CodeListAspect<TBase>.CodeList & object'.

    ../models/sap/common/index.ts:66:20:
      66 │     static readonly actions: typeof CodeList.actions & Record<neve...
         ╵                     ~~~~~~~
```
- Issue with changed function return type:
Before update:
```ts
export declare const getGoogleCalendarEvents: { (start: string, end: string): _.GoogleCalendarEvents | null, __parameters: {start: string, end: string}, __returns: _.GoogleCalendarEvents | null, kind: 'function'};
```
After update:
```ts
export declare const getGoogleCalendarEvents:  {
          // positional
          (start: string, end: string): Promise<_.GoogleCalendarEvents | null> | _.GoogleCalendarEvents | null
          // named
          ({start, end}: {start: string, end: string}): Promise<_.GoogleCalendarEvents | null> | _.GoogleCalendarEvents | null
          // metadata (do not use)
          __parameters: {start: string, end: string}, __returns: Promise<_.GoogleCalendarEvents | null> | _.GoogleCalendarEvents | null
[...]

PR #392, opened 2024-11-13
Reintroduce declare modifier on actions

daogrady:
Fixes https://github.com/cap-js/cds-typer/issues/391

We recently introduced a distinction between entities that have no actions ( = empty `.actions` property) and ones that do. In one of these branches, the required `overrides` modifier got lost. This PR streamlines both cases and reintroduces the modifier.
It also strictens the effective tsconfig that is used for transpilation tests to catch this kind of regression in the future.

### DIFF
--- a/test/tscheck.js
+++ b/test/tscheck.js
@@ -1,5 +1,12 @@
 const ts = require('typescript')
 
+const defaultTranspilationOptions = {
+    noEmit: true,
+    esModuleInterop: true,
+    strict: true,
+    noImplicitOverride: true
+}
+
 /**
  * Checks a parsed TS program for error diagnostics.
  * @param {any} program - the parsed TS program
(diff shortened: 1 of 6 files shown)


test/unit/files/actions/model.cds
```diff
@@ -85,3 +85,13 @@ service S2 {
     action a1 (p1: String, p2: many ExternalType2) returns ExternalType;
     action a2 (p1: String not null, p2: many ExternalType2, p3: Integer not null) returns ExternalType
 }
+
+entity Parent {}
```
daogrady, line 89: these new entities received no explicit test, but will cause an error during the smoke transpilation tests if the `override` modifier should go missing again.

test/tscheck.js
```diff
@@ -1,5 +1,12 @@
 const ts = require('typescript')
 
+const defaultTranspilationOptions = {
+    noEmit: true,
+    esModuleInterop: true,
+    strict: true,
+    noImplicitOverride: true
```
daogrady, line 7: this is the newly added option for transpilation tests

chgeo (approved):

PR #398, opened 2024-11-18
Add config option to generate ESM

daogrady:
cds-typer currently emits _index.js_ files that are CommonJS compliant.
Setting `targetModuleType = 'esm'` now generates ESM compliant _index.js_ files. Setting this option to `'cjs'` retains the old behaviour.

Both variants are tested by an additional test suit that runs all unit tests and tries to parse the resulting _index.js_ files.

~~Automatic project nature detection is not included in this PR. The cds-typer config is currently finalised before cds-typer actually runs. To determine a project root, we would need to have the targeted files ready, which are only passed in once cds-typer runs.~~

The default value for `targetModuleType` is `auto`, which will check `cds.root` and its parent directories for a _package.json_ to derive the `type` from.


CHANGELOG.md
```diff
@@ -6,10 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## Version 0.29.0 - tbd
 ### Added
+- Passing `--targetModuleType esm` now generates ESM compliant _index.js_ files.
 
 ### Changed
 
 ### Fixed
+- The static `.keys` property now properly reels in key types from inherited classes.
```
daogrady, line 14: supplement for https://github.com/cap-js/cds-typer/pull/394

chgeo:
> Automatic project nature detection is not included in this PR. The cds-typer config is currently finalised before cds-typer actually runs. To determine a project root, we would need to have the targeted files ready, which are only passed in once cds-typer runs.

Don't quite get this.  `package.json` is available statically, so isn't it possible to have something like `typerCfg.targetModuleType ?? require('package.json').type` ?

daogrady:
Thanks for pointing out the static way to find the project root. As per our call, the new `targetModuleType` option now also accepts `auto`, which is the default, and will check `cds.root` and its parent directories for a _package.json_ to read from.

chgeo (approved):

PR #403, opened 2024-11-20
feat: update typer schema with target_module_type

stockbal:
@daogrady should be included before updating to 0.29.0


daogrady:
Good catch! The schema definition totally slipped my mind. Thanks for adding it in! 🙂

daogrady (approved):

PR #416, opened 2024-12-02
Read default output @cds-models from package.json

daogrady:
Change default value for `outputDirectory` from `./` to `./@cds-models` when executed in a CAP project.

This PR does _not_ include changing the default for the target model files to `"*"`.


Akatuoro (approved):
Works as expected (cli option > project package.json > default) 👍

chgeo (approved):

Issue #410, opened 2024-12-02
[BUG] cds-typer error with Attachments aspect.


vc10812:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

When using the Attachments plugin ([[cap-js/attachments: The @cap-js/attachments package is a CDS plugin that provides out-of-the box asset storage and handling by using an aspect Attachments. It also provides a CAP-level, easy to use integration of the SAP Object Store.](https://github.com/cap-js/attachments#readme)](https://github.com/cap-js/attachments#readme)) for SAP CAP, an error occurs with cds typer.

I contacted the Attachments developer and they confirmed there is an issue with cds typer.
Here is the link to the Issue:
https://github.com/cap-js/attachments/issues/97#issue-2474694865

Could you please review the content?

### Environment

```markdown
- **OS**:
- **Node**:
- **npm**:
- **cds-typer**: 
- **cds**:
```

daogrady:
Hi @vc10812 ,

thanks for reporting this issue! It looks like we do not generate plural forms of aspects as of today. I will look into the matter.

Best,
Daniel

daogrady:
Hi @vc10812  ,

I have prepared a [fix](https://github.com/cap-js/cds-typer/pull/421) for the issue.
Feel free to [try it out](https://github.com/cap-js/cds-typer/wiki/Trying-Out-Prerelease-Versions) before I merge and release it.

Best,
Daniel

PR #421, opened 2024-12-04
Generate plural types for cds aspects

daogrady:
Fixes https://github.com/cap-js/cds-typer/issues/410


vc10812 (approved):
I have confirmed that the build process now works without any issues when using @cap-js/attachments.

PR #428, opened 2024-12-06
ci(release): use sed for changelog manipulation

joergmann:


.github/workflows/release.yml
```diff
@@ -60,31 +60,15 @@ jobs:
...
-            if (process.env.INPUTS_DRY_RUN === 'true') {
-              core.info(newContent);
-            }
-          } else {
-            throw 'Pattern not found in CHANGELOG.md. Manually validate changelog.';
-          }
+        NEW_VERSION: ${{ steps.get-version.outputs.new-version }}
+      run: |
+        CURRENT_DATE=$(date +%Y-%m-%d)
+        sed -i -E "0,/^##\s+\[Unreleased\]\s*$/s//## [${NEW_VERSION}] - ${CURRENT_DATE}/" CHANGELOG.md
```
daogrady, line 71:
```suggestion
        sed -i -E "0,/^##\s+\[Unreleased\].*$/s//## [${NEW_VERSION}] - ${CURRENT_DATE}/" CHANGELOG.md
```
the line we are targeting should _not_ have been modified manually. But I wouldn't be surprised if we found someone leaving a comment in this particular line at some point, i.e.

```md
## [Unreleased] (will be released after cds-dk x.z.y)
```

matching `.*` instead of `\s*` would catch this case.

daogrady (approved):
Thank you for this contribution! Looks good to me and I am happy with the action getting shorter. I left one remark, which you can consider [:fog: dust](https://gist.github.com/timtyrrell/da5aa2e7ea14230a6528777866723dd3).

Issue #405, opened 2024-11-22
[BUG] cds build fails with cds typer


qby-ankul:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

cds build throws an error for the cds typer. Furthermore, the resources are also generated at root level with the current version. Example project is a freshly generated one.

https://github.com/qby-ankul/cds-typer-assoc/blob/main/buildError.txt

### Expected Behavior

No error on build and working association.

### Steps To Reproduce

`cds build` in repo

### Environment

```markdown
BAS

user: typer-test $ cds -v

@cap-js/asyncapi: 1.0.2
@cap-js/cds-typer: 0.29.0
@cap-js/cds-types: 0.7.0
@cap-js/db-service: 1.15.1
@cap-js/openapi: 1.0.7
@cap-js/sqlite: 1.7.7
@sap/cds: 8.4.2
@sap/cds-compiler: 5.4.4
@sap/cds-dk: 8.4.2
@sap/cds-dk (global): 8.4.2
@sap/cds-fiori: 1.2.7
@sap/cds-foss: 5.0.1
@sap/cds-mtxs: 2.3.1
@sap/eslint-plugin-cds: 3.1.2
Node.js: v20.12.0
home: /home/user/projects/typer-test/node_modules/@sap/cds
typer-test: 1.0.0
```

### Repository Containing a Minimal Reproducible Example

https://github.com/qby-ankul/cds-typer-assoc/

YoussefBoudiba:
Having the same issue since v0.29.0

stockbal:
It seams the new option `target_module_type` causes an early initialization of the Configuration module:

https://github.com/cap-js/cds-typer/blob/30e777d2f08af61d87261ed4faf777ee7f3d9a5d/lib/file.js#L149-L151

which in turn makes the following lines have no effect and the `outputDirectory` remains at its default `.` instead of the automatic determined folder `@cds-models`

https://github.com/cap-js/cds-typer/blob/30e777d2f08af61d87261ed4faf777ee7f3d9a5d/cds-plugin.js#L97-L98

You can work around that problem by specifying the output directory in the package.json 

```json
{
  ...
  "cds": {
    "typer": {
      "output_directory": "@cds-models"
    }
  }
  ...
}
```

daogrady:
Hi everyone,

thanks for reporting this issue! I will push out a fix-release asap to address this problem.
[...]

PR #448, opened 2024-12-19
fix inlineDeclarations default bug

### LINKED ISSUE #447: Bug in help command lead to problems in cds build
Default values are different:
https://github.com/cap-js/cds-typer/blob/bcb1181d3b2a0abc057be61ea64eb88f1a1da982/lib/visitor.js#L65-L68
https://github.com/cap-js/cds-typer/blob/bcb1181d3b2a0abc057be61ea64eb88f1a1da982/lib/cli.js#L182-L186

kulliax:
fix #447


daogrady:
Thank you for providing a PR.
I am actually unsure if we shouldn't instead just adjust the help text, seeing that changing the default value is a breaking change.

kulliax:
Hi @daogrady 
I can't answer that. The error came with the last update, before that we also used the structured type as standard without specification. I didn't realise that this could be controlled in the options. If I see this correctly, it has been introduced here:
https://github.com/cap-js/cds-typer/blob/bcb1181d3b2a0abc057be61ea64eb88f1a1da982/CHANGELOG.md?plain=1#L26
Previously we were on 0.29.0, see: https://github.com/cap-js/cds-typer/issues/405

daogrady:
I see, is the behaviour of cds-typer now as you would expect it when you use your fix?

kulliax:
In our last build, we set the option in package.json to structured. It then ran successfully. I am open whether you change this in the help or in the default values. Perhaps you should check again in which version the standard was changed from structured to flat.

kulliax:
https://github.com/cap-js/cds-typer/commit/43d6269e6de23bfe9e3b5a88042562807dd9b6bb

as I can see, it has been flat since at least september. then the value in the help is the better option. We didn't adjust the dependencies for a while because they were causing other problems.

Thanks for your efforts!

daogrady:
Hi @kulliax ,

thanks for reporting back and for providing the fix. 🙂

Best,
Daniel

PR #487, opened 2025-02-05
Add flag to explicitly disable the typescript build task

daogrady:
Setting `cds.env.typer.skip_build_task` to `true` will now disable the `typescript` build task.
Currently, the only way to achieve this is to explicitly call `cds build --for x --for y --for z` with all build tasks, except for `typescript`.
This was requested by users who would like to perform the required build steps themselves.
PR is best viewed with hidden whitespace changes.

### DIFF
--- a/cds-plugin.js
+++ b/cds-plugin.js
@@ -52,93 +52,101 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
     )
     : undefined
 
-// FIXME: remove once cds7 has been phased out
-if (!cds?.version || cds.version < '8.0.0') {
-    DEBUG?.('typescript build task requires @sap/cds-dk version >= 8.0.0, skipping registration')
-    return
-}
+// IIFE to be able to return early
+;(() => {
+    // FIXME: remove once cds7 has been phased out
+    if (!cds?.version || cds.version < '8.0.0') {
+        DEBUG?.('typescript build task requires @sap/cds-dk version >= 8.0.0, skipping registration')
+        return
+    }
 
-// requires @sap/cds-dk version >= 7.5.0
-cds.build?.register?.('typescript', class extends cds.build.Plugin {
-    static taskDefaults = { src: '.' }
-    static hasTask() { return isTypeScriptProject() }
+    if (cds.env.typer?.skipBuildTask) {
+        DEBUG?.('skipping typescript build task registration based on configuration option')
+        return
+    }
 
-    // lower priority than the nodejs task
-    get priority() { return -1 }
+    // requires @sap/cds-dk version >= 7.5.0
+    cds.build?.register?.('typescript', class extends cds.build.Plugin {
+        static taskDefaults = { src: '.' }
+        static hasTask() { return isTypeScriptProject() }
 
-    get #appFolder () { return cds?.env?.folders?.app ?? 'app' }
+        // lower priority than the nodejs task
+        get priority() { return -1 }
 
-    /**
-     * cds.env > tsconfig.compilerOptions.paths > '@cds-models' (default)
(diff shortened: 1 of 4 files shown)


cds-plugin.js
```diff
@@ -52,93 +52,101 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
...
+        DEBUG?.('typescript build task requires @sap/cds-dk version >= 8.0.0, skipping registration')
+        return
     }
 
-    init() {
-        this.task.dest = path.join(cds.root, cds.env.build.target, cds.env.folders.srv)
+    if (cds.env.typer?.skipBuildTask) {
+        DEBUG?.('skipping typescript build task registration based on configuration option')
+        return
     }
```
daogrady, line 66: Changes in here look way extensive than they actually are, as I have now wrapped the entire registration process in an IIFE to avoid having a top level `return` for early exits.
The marked part above this comment is the actual change.

package.json
```diff
@@ -110,7 +110,7 @@
                 "flat",
                 "structured"
               ],
-              "default": "structured"
+              "default": "flat"
```
daogrady, line 113: this was changed a while ago, but not propagated to the schema in _package.json_.

daogrady:
Sure! I'd like to retain the boolean nature of the option though. The binary flags we have so far 
Is `doBuildTask` better?

chgeo:
> Is `doBuildTask` better?

No really.

I like 'terms' as option names (_build_task_ or even _build_) as they are open for becoming a config object in the future.  So:
```jsonc
{ build: true }
```
can become:
```jsonc
{ build: {
  option1: ...,
  option2: ...,
}}
```
Quite a common pattern out there, see [Vitepress example](https://vitepress.dev/reference/site-config#ignoredeadlinks).

> Sure! I'd like to retain the boolean nature of the option in the name though. 

The boolean nature can be expressed in the schema.

daogrady:
I see, that is a valid objection then. I have changed the name accordingly.

package.json
```diff
@@ -131,6 +132,11 @@
                 "auto"
               ],
               "default": "auto"
+            },
+            "build_task": {
+              "type": "boolean",
+              "description": "If set to true, the typescript build task will not be registered/ executed.\nThis value must be set in cds.env.\nPassing it as parameter to the cds-typer CLI has no effect.",
```
chgeo, line 138:
```suggestion
              "description": "If set to true, the typescript build task will not be registered/executed.\nThis value must be set in package.json.\nPassing it as parameter to the cds-typer CLI has no effect.",
```

daogrady: What about the other ways to write to `cds.env`, like cdsrc.json, env.properties, etc?

chgeo: Yes, these could also be valid, mentioning is ok here for brevity I'd say.  In contrast, people would not know how to write to cds.env.  Where is cds.env on my machine? ;)

daogrady: I would have hoped for anyone not knowing to go to CAPire and (hopefully) be sent to appropriate articles
Screenshot 2025-02-06 at 13 56 26

You probably have more insight into this than I do: do users being asked to "set it in package.json" generally know how to set it there? If we usually tell people to "set this in your package.json" and they know they have to add it to the `"cds:"` section there, then I am perfectly fine phrasing it that way.

chgeo: Yeah, some people know what cds.env is, esp. Node people.  But newbies don't.  This is why we added the config popup in capire to be able to display all config options.

I'd be fine with _set it your project configuration_.

daogrady: That's a good compromise :)

chgeo (approved):

PR #488, opened 2025-02-06
Add schema consistency check

daogrady:
cds-typer supports a bunch of [CLI parameters](https://github.com/cap-js/cds-typer/blob/main/lib/cli.js#L146) that can all also be specified via `cds.env`.
To receive code completion for these options when editing a project's _package.json_, these options, default values, and descriptions also have to be maintained [in cds-typer's _package.json_](https://github.com/cap-js/cds-typer/blob/main/package.json#L69).
The latter can be generated automatically using [a utility script](https://github.com/cap-js/cds-typer/blob/main/scripts/write-cds-typer-schema.js). But if anyone adds a new configuration option and forgets to run this script, the code completion will be incorrect.
The workflow introduced in this PR therefore blocks any PR where the author apparently forgot to run the script.

### DIFF
--- a/.github/workflows/check-schema-integrity.yml
+++ b/.github/workflows/check-schema-integrity.yml
@@ -30,6 +30,8 @@ jobs:
           git config --global user.email 'github-actions[bot]@users.noreply.github.com'
           git add .
           if ! git diff --cached --exit-code; then
-            echo "Schema in package.json not consistent with CLI parameters. Run `npm run write:cds-typer-shema` and commit the resulting changes."
+            echo "Schema in package.json not consistent with CLI parameters. Run "npm run write:cds-typer-shema" and commit the resulting changes."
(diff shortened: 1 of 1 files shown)


daogrady:
The option this [PR is currently failing on](https://github.com/cap-js/cds-typer/pull/487/files#r1943252890) is the reason for this PR. It will go away once that PR is merged and this PR is rebased against `main`.

joergmann:
lgtm, another option would be to add a commit which the updated schema. Even if user forgets, he will get a zwangscommit with the schema into his PR

daogrady:
I had considered that, but I am always super cautious with automatic commits to human PRs. I guess we could have automatically stacked another PR with the required change on top, but I hope we won't see too many changes to the CLI options in the long run.

joergmann (approved):

PR #506, opened 2025-03-04
Run typer once before cds watch starts

daogrady:
Running `cds w` in your project will now initially run `cds-typer` if no `@cds-models` folder (or to whatever it is configured via `cds.typer.output_directory`) doesn't exist.
It does not run if this directory exists.
This feature is especially useful when you check out a CDS project and immediately dive into `cds w` from the CLI.

### DIFF
--- a/cds-plugin.js
+++ b/cds-plugin.js
@@ -159,10 +159,10 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
 if (cds.watched) {
     if (fs.existsSync(cds.env.typer.output_directory)) return
     try {
-        DEBUG?.('>> running cds-typer before cds watch')
+        DEBUG?.('>> start cds-typer before cds watch')
         module.exports = typer.compileFromFile('*')
-            .then(() => DEBUG?.('<< running cds-typer before cds watch')
-        )
+            .then(() => DEBUG?.('<< end cds-typer before cds watch'))
+            .catch(e => DEBUG?.(e))
     } catch (error) {
         // silently ignore if no (valid) model exists at initial startup
         DEBUG?.(error)
--- a/cds-plugin.js
+++ b/cds-plugin.js
@@ -55,7 +55,7 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
     )
     : undefined
 
-cds.on('on-watch', async () => {
+cds.on('watch', async () => {
     if (fs.existsSync(cds.env.typer.output_directory ?? DEFAULT_MODEL_DIRECTORY_NAME)) return
     try {
         DEBUG?.('running cds-typer before cds watch')
--- a/cds-plugin.js
+++ b/cds-plugin.js
@@ -55,16 +55,6 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
     )
     : undefined
 
-cds.on('watch', async () => {
-    if (fs.existsSync(cds.env.typer.output_directory ?? DEFAULT_MODEL_DIRECTORY_NAME)) return
-    try {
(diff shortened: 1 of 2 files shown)


cds-plugin.js
```diff
@@ -52,6 +55,16 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
     )
     : undefined
 
+cds.on('on-watch', async () => {
```
chgeo, line 58: The event name sounds redundant.
```suggestion
cds.on('watch', async () => {
```

cds-plugin.js
```diff
@@ -52,6 +55,16 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
...
+    if (fs.existsSync(cds.env.typer.output_directory ?? DEFAULT_MODEL_DIRECTORY_NAME)) return
+    try {
+        DEBUG?.('running cds-typer before cds watch')
+        await typer.compileFromFile('*')
+    } catch  {
+        // silently ignore if no (valid) model exists at initial startup
```
chgeo, line 64: Maybe log the error with `DEBUG` or so?

cds-plugin.js
```diff
@@ -52,6 +55,16 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
     )
     : undefined
 
+cds.on('on-watch', async () => {
+    if (fs.existsSync(cds.env.typer.output_directory ?? DEFAULT_MODEL_DIRECTORY_NAME)) return
```
chgeo, line 59: Shouldn't `cds.env.typer.output_directory` always default to `@cds-models`? 
See https://github.com/cap-js/cds-typer/blob/d73780c40b5d017194fa9c8475f7f72318674c33/package.json#L57

No other fall-fallback needed then.

cds-plugin.js
```diff
@@ -5,9 +5,12 @@ const util = require('util')
 const exec = util.promisify(require('child_process').exec)
 const typer = require('./lib/compile')
 const { fs, path } = cds.utils
+const { configuration } = require('./lib/config')
+
 const DEBUG = cds.debug('cli|build')
 const BUILD_CONFIG = 'tsconfig.cdsbuild.json'
-const { configuration } = require('./lib/config')
+const DEFAULT_MODEL_DIRECTORY_NAME = '@cds-models'
```
chgeo, line 12: Seems redundant to me.  See comment below.

cds-plugin.js
daogrady: It should, yes. I am a bit paranoid about it being unset for whatever reason, but we can forgo this.

chgeo: I know that feeling :)
But at some point we have to trust the config system, otherwise this leads to duplication of default values all over the place.

cds-plugin.js
```diff
@@ -151,4 +154,17 @@ const rmFiles = async (dir, exts) => fs.existsSync(dir)
...
+    try {
+        DEBUG?.('>> running cds-typer before cds watch')
+        module.exports = typer.compileFromFile('*')
+            .then(() => DEBUG?.('<< running cds-typer before cds watch')
+        )
+    } catch (error) {
+        // silently ignore if no (valid) model exists at initial startup
+        DEBUG?.(error)
+    }
+}
```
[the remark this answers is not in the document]
chgeo: > * run once on the initial startup of `cds w`

Yes.  Just give it a try :)

> * honour a passed project parameter, i.e. `cds w bookshop`

This is the feature of the plugin system that I mentioned above: it awaits any exported Promise.

chgeo (approved):

PR #510, opened 2025-03-17
Use true as default for use_entities_proxy

daogrady:
`cds.typer.use_entities_proxy = true` is now default behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 ### Changed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 ### Changed
+- [breaking] The config `use_entities_proxy` is now set to `true` as default. You can set it to `false` to revert to the old behaviour.
```
chgeo, line 9:
```suggestion
- [breaking] The config `use_entities_proxy` (allowing for static imports) is now set to `true` as default. You can set it to `false` to revert to the old behaviour.
```

chgeo (approved):

PR #513, opened 2025-03-19
Add hashing

daogrady:
Setting `cds.typer.cache = 'blake2s256'` hashes the extended model and compares it to the hash of the previous run read from disk (if it exists). If the model has not changed, cds-typer will end the run immediately at this point.

cds-typer currently needs the inferred and xtended flavour of a model for a complete run. So when actually running, it performs `cds.load` twice, considerably increasing runtime.
With the `cache` setting active, it loads the xtended model first, checks the hash, and if the hash has not changed, does not perform the second `cds.load`. So we should see a runtime improvement of about 50% for the initial model loading when the cache is hit.

File IO is also cut down in case of a cache hit, as no files have to be written.

A quick test with a model of ~8MB in size showed that the time went from ~13s on the initial run when no `@cds-models` existed, down to ~5s when cds-typer would stop immediately after finding there was nothing to do.

```
➜ time cds-typer "*"
cds-typer "*"  13.03s user 1.12s system 133% cpu 10.613 total
➜ time cds-typer "*"
cds-typer "*"  5.35s user 0.32s system 136% cpu 4.159 total
```

The currently used algorithm [blake2](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2) is, to my knowledge, the fastest hash algorithm available in the `crypto` module. I deliberately left the config option open to using a different algorithm in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- Introduced `cache` option to cache CDS model between runs. If set to `blake2s256`, typer will only regenerate the files if the model has changed between runs.
 ### Changed
 - [breaking] The config `use_entities_proxy` (allowing for static imports) is now set to `true` as default. You can set it to `false` to revert to the old behaviour.
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
-- Introduced `cache` option to cache CDS model between runs. If set to `blake2s256`, typer will only regenerate the files if the model has changed between runs.
+- Improved performance of repeated `cds-typer` runs by introducing an option to cache CDS model between runs. By default, typer will only regenerate the files if the model has changed between runs, using `blake2s256` as algorithm.
 ### Changed
 - [breaking] The config `use_entities_proxy` (allowing for static imports) is now set to `true` as default. You can set it to `false` to revert to the old behaviour.
 ### Deprecated
--- a/lib/compile.js
+++ b/lib/compile.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const fs = require('fs')
-const { normalize } = require('path')
+const { normalize, join } = require('path')
 const cds = require(require.resolve('@sap/cds', { paths: [process.cwd(), __dirname] }))
 const util = require('./util')
 const { writeout } = require('./file')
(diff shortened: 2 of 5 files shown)


lib/compile.js
```diff
@@ -55,10 +75,31 @@ const compileFromCSN = async csn => {
  * @param {string | string[]} inputFile - path to input .cds file
  */
 const compileFromFile = async inputFile => {
+    const hashFile = join(configuration.outputDirectory, '.hash')
     const paths = typeof inputFile === 'string' ? normalize(inputFile) : inputFile.map(f => normalize(f))
-    const xtended = await cds.linked(await cds.load(paths, { docs: true, flavor: 'xtended' }))
+    const xtendedPlain = await cds.load(paths, { docs: true, flavor: 'xtended' })
+    let modelHash
+
+    if (configuration.cache === 'blake2s256') {
```
chgeo, line 83: Why not just accept any hash algorithm here, so any value from `cds.typer.hash` except `none`?
Would of course only make sense if there are other reasonable algorithms.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- Introduced `cache` option to cache CDS model between runs. If set to `blake2s256`, typer will only regenerate the files if the model has changed between runs.
```
chgeo, line 8:
```suggestion
- Improved performance of repeated `cds-typer` runs by introducing an option to cache CDS model between runs. By default, typer will only regenerate the files if the model has changed between runs, using `blake2s256` as algorithm.
```

lib/compile.js
daogrady: I don't see a reason for allowing arbitrary algorithms here: blake2 [is faster](https://www.blake2.net/) than the other hash algorithms included in `node:crypto`, so users should not choose a slower one, like SHA.
On the other hand, there _are_ faster hash algorithms, like [xxhash or Murmur](https://github.com/Cyan4973/xxHash#benchmarks), but they would require an external dependency. Should we ever decide to include this dependency, or if `node:crypto` introduces more algorithms, I'd like to be able to allow a new value for `cache`, so users can migrate to new algorithms while maintaining backwards compatibility. I think giving the user the option to pick any algorithm could be overwhelming, if they do not want to evaluate the pros and cons of each option. So allowing a narrow, but expandable, set of values should provide gentle guidance.

chgeo: Ok, agreed.

chgeo (approved):

hakimio:
@daogrady it might be a good idea to include `cds-typer` version in the hash since output changes from one `cds-typer` version to another.

daogrady:
Hi Tomas,

that is a good point! I'll follow up on that.

Best,
Daniel

PR #520, opened 2025-03-31
chore: prep cds9

johannes-vogel:


daogrady (approved):

daogrady:
Thanks for the PR! I removed cds from `devDependencies` completely, as we have an open range for cds in the `peerDependencies` already. This should have the same effect and come with less maintainance.

Issue #547, opened 2025-06-12
[BUG] TypeError: Cannot set properties of undefined (setting 'notNull') during cds build


codeyogi911:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

The `@cap-js/cds-typer` package is throwing a TypeError during the `cds build` process when processing **consumption service definitions** and **consumption model entities** that use **composition of aspects**. The build completes successfully but the TypeScript type generation fails with a null reference error specifically when processing aspect composition patterns.

## Error Message

```
TypeError: Cannot set properties of undefined (setting 'notNull')
    at Resolver.resolveAndRequire (/Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/resolution/resolver.js:322:32)
    at Resolver.resolveAndRequire (/Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/resolution/resolver.js:322:32)
    at FlatInlineDeclarationResolver.visitElement (/Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/components/inline.js:88:44)
    at FlatInlineDeclarationResolver.visitElement (/Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/components/inline.js:88:44)
    at Visitor.visitElement (/Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/visitor.js:625:47)
    at /Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/visitor.js:240:26
    at Buffer.addIndented (/Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/file.js:692:13)
    at Buffer.addIndentedBlock (/Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/file.js:709:14)
    at /Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/visitor.js:232:20
    at Buffer.addIndented (/Users/shashwatjain/Repos/mymediset_cloud/node_modules/@cap-js/cds-typer/lib/file.js:692:13)
[...]

PR #548, opened 2025-06-12
Resolve composition aspect targets

daogrady:
Fixes https://github.com/cap-js/cds-typer/issues/547

Compositions targeting aspects, instead of entities, would carry this information in a property `targetAspect`, rather than `target`. This PR falls back to resolving `targetAspect` when `target` can not be resolved.


daogrady:
> Add a test?

Test has been added

Issue #545, opened 2025-06-10
[BUG] Declare statements missing from type definitions used in entities


MarvinWeitz:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

When defining a type inside a `cds` file, the generated types are missing a declare statement.

```
type geoCoordinates : {
  latitude  : Decimal(10, 6) @Common.Label: '{i18n>Latitude}';
  longitude : Decimal(10, 6) @Common.Label: '{i18n>Longitude}';
}
```

will result in 

```
return class geoCoordinates extends Base {
  latitude: number | null
  longitude: number | null
  static readonly kind: 'entity' | 'type' | 'aspect' = 'type';
  declare static readonly keys: __.KeysOf<geoCoordinates>;
  declare static readonly elements: __.ElementsOf<geoCoordinates>;
  declare static readonly actions: globalThis.Record<never, never>;
};
```

### Expected Behavior

The expected result should be

```
return class geoCoordinates extends Base {
  declare latitude: number | null
  declare longitude: number | null
  static readonly kind: 'entity' | 'type' | 'aspect' = 'type';
  declare static readonly keys: __.KeysOf<geoCoordinates>;
  declare static readonly elements: __.ElementsOf<geoCoordinates>;
  declare static readonly actions: globalThis.Record<never, never>;
};
```

### Steps To Reproduce

1. cds init
2. cds add typescript
3. npm i
4. Create type as above

### Environment

```markdown
- **OS**: Debian GNU/Linux 12 (bookworm)
- **Node**: v20.18.1
- **npm**: 11.4.1
- **cds-typer**: 0.35.0
- **cds**: 8.9.1
```

### Anything else?

I'll try to fix this as my first PR in the coming week / weekend, if my time schedule allows for it!

MarvinWeitz:
@daogrady I finally had some time to address this issue and found a solution. How exactly do I go about adding the changes in a PR? I don't have the permissions to create a new branch.

daogrady:
Hi Marvin,

glad you found the time! You can [fork the repository](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo) and [open a repository from there](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork).

Best,
Daniel

[...]

PR #554, opened 2025-06-30
Fix declare statement for type definitions while optional properties are disabled

MarvinWeitz:
Fix for #545.
- Changes the check on adding a declare modifier to properly handle type definitions
- Sets properties as non optional in the transpilation test to avoid future errors

### DIFF
--- a/lib/components/property.js
+++ b/lib/components/property.js
@@ -1,10 +1,12 @@
 /**
  * Determines the proper modifiers for a property.
- * For most properties, this will be "declare". But for some properties,
- * like fields of a type, we don't want any modifiers.
- * A check for the parents name is necessary to avoid adding declare to
- * properties of a object type reference. These will have the object type as
- * their parent's name.
+ * 
+ * All properties must be modified by `declare`.
+ * The only exception to this is in case of an action return type being defined
+ * as an object. Here `declare` would cause an error.
+ * This case can be identified by checking the parent of the property:
+ * If the parent is a type and has no name, then it is an action return type.
+ * 
  * @param {import('../typedefs').resolver.EntityCSN} element - The element to determine the modifiers for.
  * @returns {import('../typedefs').resolver.PropertyModifier[]} The modifiers for the property.
  */
--- a/lib/components/property.js
+++ b/lib/components/property.js
@@ -2,10 +2,13 @@
  * Determines the proper modifiers for a property.
  * For most properties, this will be "declare". But for some properties,
  * like fields of a type, we don't want any modifiers.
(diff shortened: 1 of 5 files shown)


lib/components/property.js
```diff
@@ -2,10 +2,13 @@
...
  * For most properties, this will be "declare". But for some properties,
  * like fields of a type, we don't want any modifiers.
+ * A check for the parents name is necessary to avoid adding declare to
+ * properties of a object type reference. These will have the object type as
+ * their parent's name.
  * @param {import('../typedefs').resolver.EntityCSN} element - The element to determine the modifiers for.
  * @returns {import('../typedefs').resolver.PropertyModifier[]} The modifiers for the property.
  */
-const getPropertyModifiers = element => element?.parent?.kind !== 'type' ? ['declare'] : []
+const getPropertyModifiers = element => element?.parent?.kind === 'type' && element?.parent?.name === undefined ? [] : ['declare']
```
daogrady, line 11: Can you please point out what CSN structure would fulfil this condition? I.e. what kind of definition in a .cds file would still lead to a `declare` modifier after this change?

daogrady:
Hi Marvin,

thank you for your contribution! Looks good overall and setting `propertiesOptional: false` for the transpilation tests only is a good idea.
See my one comment and please also add a changelog entry to satisfy the blocking PR voter. Thanks!

Best,
Daniel

lib/components/property.js
MarvinWeitz: It's better to go the other way around. This will _not_ add a `declare` modifier in case of an action return type being defined as an object. In the current test data this would be the case in the `actions/model.cds`:

``` cds
action h () returns { a: Integer; b: String };
```

This must be prevented, as the result would be

``` ts
h:  {
    // positional
    ():     {
      declare a?: number | null;
      declare b?: string | null;
    }
    ...
 }
 ```

In these cases `element.kind.parent` is `type` (which seems like the default if no other `kind` can be determined). It can only be differentiated by checking `element.parent.name` which is `undefined` in the return type case, but in the  type definition case it does have a parent name (the object type name).

Should I add a shorter version to the js docs as well or is it enough to have in the PR comment?

daogrady: If you could add this example in code, that would be greatly appreciated!
As you have probably experienced by now, we have to employ a lot of conditional logic for some very particular CDS constructs, which is not always clear on first glance, which can lead to such logic falling apart during refactoring cycles.
A comment would be nice, a test cast would be even better!

daogrady (approved):
Looks good, thanks for the contribution!

PR #556, opened 2025-07-01
Add `__self` meta information to bound actions

daogrady:
Bound actions now carry a reference to the entity they are bound to.
This can then be consumed by cds-types when inferring `req.subject`:

```ts
// in a service
this.on(myBoundAction, req => {
  req.subject  // so far this is cds.req, but could be inferred to a stricter type based on __self
})
```


daogrady:
@chgeo this is basically an experiment as of now, especially as the reporter is not certain how `req.subject` should be typed themselves. Do you have an opinion on this? We can also postpone it to the next type sync.

PR #565, opened 2025-08-08
Test/changelog check

joergmann:


daogrady:
Thanks for this contribution!
Do we still want to be able to skip the workflow, i.e. by adding a label? This wf does not seem to react to the `skip workflow` label as of now.
Seeing that this particular PR fails, I guess `chore:` or `ci:` would be the more fitting  [prefix](https://www.conventionalcommits.org/en/v1.0.0/#summary). Or did you choose `fix:` on purpose to demonstrate the workflow?

8ydfc

PR #567, opened 2025-08-28
feat: Add (experimental) CDS env `cds.env.typer.output_d_ts_files`

### LINKED ISSUE #546: [FEATURE] Generate index.d.ts files, instead of index.ts files
### Description

So far, cds-typer generates a pair of index.js and index.ts files per namespace. While the former is required to provide runtime support and make sure entities are present and available when cds is started, the latter is meant to describe the types of the entities.

The index.ts files contain fully-fledged classes, which are sometimes picked up by the TS compiler, overwriting the accompanying index.js files with incorrect information.

### Suggested Solution

Generating index.d.ts files could be an option. It would be preferable to strictly adhere to a declarative style in .d.ts files, i.e. not generate function bodies or class bodies. This can [indeed be achieved](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAwg9gOwM7AE4FcDGwCGAjAG2gF4oEIB3KACgDp6dUBzJALihwRAG0BdASijEAfFDh4AVhGwAoGZgI4kSKAFEEwAJagoAbxlRDUACbTFqaAA92AFgBMMgL5zTCxtABm6BNk2IoAEJwcADWAIJIkNgAPAAqAUrQEJbAEAjGKvDIaFi4hBDC1AlIEOzxifzs+kZklDT0tIws7Jw8AlUGNUYIOAC2EAD87CiomghMnV0mEEiYo2BaiENQI2MTNc41FjjGiAQgUCFjxuwARGlaoKdQAD5Qp6CQ13enSlHAp5NQ27sI+xy+RBsKAAJWkcFQxmi5AAbhBUAAaWpw1DCJxQABkUHKJTkMlc5mgbmUgWCISgyVS6RUQVCEXe1HUlxAgl0zgJ7igxJpZJUlLSGSgYVQqBwIGitJComqRhQOC0mG+EB2ewOABJMHBvMB2Ah0L08PCvnKFUqVX8DjhAch2GDNZDoRAUUjYfC0c55EDgFA8EJalRJdR+J7sj6VKRyAHeUGZJLGtakLHebQNVqNDI8LRTLN5osEBnaJZDAB6YtQDyoOC9NQabQgGSlqAAAWASAAtMl3h2RRCCxRK+MM0haB4IaorQALai+kQ+2g9frBoA). Having a pair of index.js and index.d.ts would keep tsc from compiling over the existing index.js file.

We need to investigate if this approach has any other implications in terms of importing them during runtime, them being present during build time, and using them during runtime.

### Alternatives

An alternative approach would be to actually have fully usable classes in the index.ts files. So even if they are compiled into index.js files, the generated artefacts can be picked up and understood by the cds framework.
This already works in parts as of today, as we add an artificial `.name` property containing the fully qualified identifier into the generated classes. The cds framework, in some places, falls back to inspecting the `.name` property and uses that to perform a lookup in the csn.
For this to work, we'd need to make sure

(a) to really have a fully qualified `.name` in every generated artefact
(b) any API of the cds framework that can come into touch with cds-typer artefacts can and will perform the lookup on the `.name` property
(c) the additional lookup does not pose an unacceptable overhead 

### Additional Context

_No response_

siarhei-murkou:
_It partially completes #546, but not completely implements the `*.d.ts` files generation from the scratch._

---

**TODO**:
- [x] Set up new _boolean-like_ CDS env `output_d_ts_files` (default: false)
- [x] Update `CHANGELOG.md`
- [x] Fix tests (no need to fix - it works as expected with disabled (default: false) option
- [x] Test out in _real_ (production) project — (`*.d.ts` files are working fine with 0 TS errors! 🤠)
- [x] Add new tests

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -150,7 +150,7 @@ const flags = enrichFlagSchema({
         type: 'string'
     },
     outputDTsFiles: {
-        desc: '(experimental) If set to true, *.d.ts files will be generated per each *.js files. If set to false (default), *.ts files will be generated per each *.js files.',
+        desc: '(experimental) If set to true, emits .d.ts files for each generated .js file. If set to false (default), emits .ts files instead.',
         default: 'false',
         type: 'boolean'
     },
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -149,6 +149,11 @@ const flags = enrichFlagSchema({
         default: './',
         type: 'string'
     },
+    outputDTsFiles: {
+        desc: '(experimental) If set to true, *.d.ts files will be generated per each *.js files. If set to false (default), *.ts files will be generated per each *.js files.',
+        default: 'false',
+        type: 'boolean'
+    },
     help: {
         desc: 'This text.',
     },
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -149,11 +149,10 @@ const flags = enrichFlagSchema({
         default: './',
         type: 'string'
     },
-    outputDTsFiles: {
-        desc: '(experimental) If set to true, emits .d.ts files for each generated .js file. If set to false (default), emits .ts files instead.',
+    outputDTsFiles: parameterTypes.boolean({
+        desc: '(experimental) If set to true, emits .d.ts files for each generated .js file. If set to false (default), emits .ts files instead. Note: skipLibCheck must be set to true in your tsconfig for this option to work properly.',
         default: 'false',
-        type: 'boolean'
-    },
+    }),
     help: {
         desc: 'This text.',
     },
--- a/lib/file.js
+++ b/lib/file.js
@@ -3,6 +3,7 @@
 const fs = require('fs').promises
 const path = require('path')
 const { readFileSync } = require('fs')
+const cds = require('@sap/cds')
 const { printEnum, propertyToInlineEnumName, stringifyEnumImplementation } = require('./components/enum')
 const { normalise } = require('./components/identifier')
 const { empty } = require('./components/typescript')
@@ -825,10 +826,11 @@ const writeout = async (root, sources) =>
     Promise.all(
         sources.map(async source => {
             const dir = path.join(root, source.path.asDirectory({local: false, posix: false}))
+            const output_file = cds.env.typer.output_d_ts_files ? 'index.d.ts' : 'index.ts'
             try {
                 await fs.mkdir(dir, { recursive: true })
                 await Promise.all([
-                    fs.writeFile(path.join(dir, 'index.ts'), source.toTypeDefs()),
+                    fs.writeFile(path.join(dir, output_file), source.toTypeDefs()),
                     fs.writeFile(path.join(dir, 'index.js'), source.toJSExports()),
                 ])
 
--- a/lib/file.js
+++ b/lib/file.js
@@ -825,7 +825,7 @@ const writeout = async (root, sources) =>
     Promise.all(
         sources.map(async source => {
             const dir = path.join(root, source.path.asDirectory({local: false, posix: false}))
-            const outputFile = configuration.outputDTSFiles ? 'index.d.ts' : 'index.ts'
+            const outputFile = configuration.outputDTsFiles ? 'index.d.ts' : 'index.ts'
             try {
                 await fs.mkdir(dir, { recursive: true })
                 await Promise.all([
--- a/lib/file.js
+++ b/lib/file.js
@@ -3,7 +3,6 @@
 const fs = require('fs').promises
 const path = require('path')
 const { readFileSync } = require('fs')
-const cds = require('@sap/cds')
 const { printEnum, propertyToInlineEnumName, stringifyEnumImplementation } = require('./components/enum')
 const { normalise } = require('./components/identifier')
 const { empty } = require('./components/typescript')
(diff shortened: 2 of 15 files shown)


siarhei-murkou:
_**TODO:** Figure out if all these tests have to have a duplicated it's per `index.ts` & `index.d.ts` or it's fine to create a small separate `describe(...)` having new tests for `index.d.ts`_

image

test/smoke/transpilation.test.js
```diff
@@ -20,62 +21,71 @@ const modelDirs = fs.readdirSync(locations.smoke.files(''))
...
-    modelDirs.forEach(({ name, rootFile }) => {
-        it(name, async () => {
-            try {
+for (const { output_d_ts_files, output_file } of [
+    { output_d_ts_files: false, output_file: 'index.ts' },
+    { output_d_ts_files: true, output_file: 'index.d.ts' },
+]) {
+    before(() => {
+        cds.env.typer.output_d_ts_files = output_d_ts_files
+    })
```
siarhei-murkou, line 30: hello @daogrady ,

please let me know if you think that covering **_"output_d_ts_files=true/false"_** in this file (for transpilation) makes no sense... it is not using `*.ts` or `*.d.ts` directly, but probably this option affects the logic inside this test as well... wdyt? 🤔

daogrady:
Hi Siarhei,

thanks for investigating the relevant option!
That does indeed open up a whole different can of worms, as it disables checking of _.d.ts_ files (and thus for the files emitted by cds-typer) entirely. That would allow:
```ts
export const x: number = "hello world"
```
without causing any errors at all. While that is nice at first glance, it may make tracking errors actually harder, as they will only become apparent down the line, and not cause transpilation errors. 🤔

I am not opposed to merging the current state, as it unblocks upgrades to typescript@5.7 and is marked as "experimental" in the option schema. I would just like to keep the linked issue #546 open for now and make the relation to `skipLibCheck` a bit more explicit.

Best,
Daniel

lib/cli.js
```diff
@@ -149,6 +149,11 @@ const flags = enrichFlagSchema({
         default: './',
         type: 'string'
     },
+    outputDTsFiles: {
+        desc: '(experimental) If set to true, emits .d.ts files for each generated .js file. If set to false (default), emits .ts files instead.',
```
daogrady, line 153:
```suggestion
        desc: '(experimental) If set to true, emits .d.ts files for each generated .js file. If set to false (default), emits .ts files instead. Note: skipLibCheck must be set to true in your tsconfig for this option to work properly.',
```

lib/file.js
```diff
@@ -825,10 +826,11 @@ const writeout = async (root, sources) =>
     Promise.all(
         sources.map(async source => {
             const dir = path.join(root, source.path.asDirectory({local: false, posix: false}))
+            const output_file = cds.env.typer.output_d_ts_files ? 'index.d.ts' : 'index.ts'
```
daogrady, line 829: Please use:
```suggestion
            const output_file = configuration.outputDTsFiles ? 'index.d.ts' : 'index.ts'
```
this mixes CLI and cds.env and allows passing the parameter as both snake_case and camelCase.

Please also use camelCase for variables (you consistently used `output_file`, it should be `outputFile`). That one is partially on me, I should have this in our eslint config.

lib/file.js
```diff
@@ -3,6 +3,7 @@
 const fs = require('fs').promises
 const path = require('path')
 const { readFileSync } = require('fs')
+const cds = require('@sap/cds')
```
daogrady, line 6:
```suggestion
const { configuration } = require('./config')
```
see below

siarhei-murkou:
> it may make tracking errors actually harder, as they will only become apparent down the line, and not cause transpilation errors

hmm... true, but isn't it _the responsibility_ of the lib itself, but not of the application installing that lib?.. 🤔 e.g. doing some `ts-check` before emitting generated TS files, but not sure how complex is to add it programmatically to the `cli.js` workflow to prevent emitting such invalid TS code lines...

> I am not opposed to merging the current state, as it unblocks upgrades to typescript@5.7 and is marked as "experimental" in the option schema. I would just like to keep the linked issue https://github.com/cap-js/cds-typer/issues/546 open for now and make the relation to skipLibCheck a bit more explicit.

ok — "unlinked" 🙂

lib/file.js
daogrady: I have added the camelcase rule to the eslint config, if you merge in `main`, it should hopefully help finding the snake_case variables/ parameters.

siarhei-murkou: should _camelCase_ be used for the object _properties_ too? e.g. here -> [test.config.js#R11-R12](https://github.com/cap-js/cds-typer/pull/567/files#diff-324e5ae6ed1d64846d943bc35a8a7394f2099af356ca5d9e30413648a3ae124fR11-R12)

secondly, if you missed it could you please leave any comment / opinion on [this change](https://github.com/cap-js/cds-typer/pull/567#discussion_r2355548287) for the smoke test?.. 🙂

siarhei-murkou: right, but wanted to do smth _more compact, less noisy_ 😅

daogrady: so you'd use the `options` parameter and access the properties thereof instead of destructuring it? I don't see a problem with that, and I don't think it's a hack. Or am I missing something?

daogrady: I am unclear on why you use the snake_case style in your `const CDS_TEST_CONFIG_OPTIONS`to begin with. I have not seen any place where just using 

```js
const CDS_TEST_CONFIG_OPTIONS = [
    { outputDTsFiles: false, outputFile: 'index.ts' },
    { outputDTsFiles: true, outputFile: 'index.d.ts' },
]
```
would be a problem 🤔

daogrady (approved):
Looks good now, thanks for the contribution!

PR #592, opened 2026-02-16
feat: added support for multi-project setups

Vo1dSh4d0w:
Hello,

these changes allow multi-project setups to be compiled using cds-typer. We need this change, because our project structure does not exactly match the structure in our target directory (gen), therefore we have a tsconfig each for our 2 separated projects.

With these changes, in order to build 2 projects, the build defintion has to be modified manually, for example:
```json
  "build": {
    "target": "gen",
    "tasks": [
      {
        "for": "hana",
        "src": "db"
      },
      {
        "for": "nodejs",
        "src": "srv"
      },
      {
        "for": "mtx-sidecar",
        "src": "mtx/sidecar"
      },
      { "for": "typescript" },
      { "for": "typescript", "src": "mtx/sidecar", "dest": "mtx/sidecar" }
    ]
  }
```

This change works by joining every project related path (like tsconfig.cdsbuild.json) with the configured "src", I've also added a check to the configured "dest", which ensures that a custom "dest" won't be overwritten in the "init()" function.

I've tested this patch both with our real project and with a smaller project without any explicit build configuration.

### DIFF
--- a/cds-plugin.js
+++ b/cds-plugin.js
@@ -23,10 +23,18 @@ const isTypeScriptProject = () => {
 }
 
 /**
+ * Get the expected location of the tsconfig file with respect for a custom project root.
+ * @param {string} src - The project root
+ * @returns {string}
+ */
+const buildConfigLocation = (src) => path.join(src, BUILD_CONFIG)
+
+/**
  * Check if separate tsconfig file that is used for building the project.
+ * @param {string} src - The project root
  * @returns {boolean}
  */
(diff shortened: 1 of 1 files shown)


daogrady:
Hi Filip,

thanks for pointing this scenario out and for providing a fix.
The change looks good on first glance, please just address the failing workflows.

Also, could you please let me know why your projects need differing layouts? This is a use case we have not yet encountered.

Best,
Daniel

PR #607, opened 2026-04-21
fix: export inline enums so they are accessible cross-namespace in service projections

### LINKED ISSUE #518: [BUG] Association key not correctly typed when String with asserted enum range
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Nature of Your Project

TypeScript

### Current Behavior

When having an entity that has a key of type `String @assert.range enum { ... }`

```
entity BookIDs {
  key ID : String @assert.range enum {
        ID1;
        ID2;
      };
}
```

and this entity is used in an Association of another entity

```
entity Books {
  key ID    : Association to BookIDs;
      title : String;
      stock : Integer;
}
```

which in tern is used in a service

```
service CatalogService {
    @readonly
    entity Books as projection on my.Books;
}
```

the generated types break, as the association key of `Books` in the service definition uses the custom type of `BookID_ID` which is neither defined locally nor imported from the entity ts file.

### Expected Behavior

The type is either imported or generated in file.

### Steps To Reproduce

1. `cds init`
1. `cds add typescript`
1. `npm i`
1. Create entities as above (or see repo below)
1. Look at `@cds-models/CatalogService/index.ts`

### Environment

```markdown
- **OS**: Debian GNU/Linux 12 (bookworm)
- **Node**: v20.18.1
- **npm**: 11.1.0
- **cds-typer**: 0.33.1
- **cds**: 8.8.3
```

### Repository Containing a Minimal Reproducible Example

https://github.com/MarvinWeitz/cds-typer-service-enum-bug

### Anything else?

I guess the problem will come with any specially typed key used this way, I just encountered it with enum asserted strings.

MarvinWeitz:
When an entity has a key with an inline enum type (e.g. `key ID : String enum { ID1; ID2; }`) and that entity is the target of an `Association` in another entity that is then projected in a service, the generated code breaks.

The service file correctly received a namespaced reference like `_my_bookshop.BookID_ID` for the foreign key type, but `BookID_ID` was never exported from `my/bookshop/index.ts` so the reference resolved to `undefined` at runtime and produced a TypeScript compile error

Closes #518

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 ### Deprecated
 ### Removed
 ### Fixed
+- Association keys pointing to entities with inline enum types are now correctly typed in service projections across namespace boundaries
 ### Security
 
 ## [0.38.0] - 2025-09-24
--- a/lib/file.js
+++ b/lib/file.js
@@ -326,9 +326,8 @@ class SourceFile extends File {
         entityProxy.push(propertyName)
 
         // REVISIT: find a better way to do this???
-        const printEnumToBuffer = (/** @type {Buffer} */buffer) => printEnum(buffer, propertyToInlineEnumName(entityCleanName, propertyName), kvs, {export: false}, doc)
-
         if (buffer?.namespace) {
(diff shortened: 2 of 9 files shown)


daogrady (approved):
Hi Marvin,

thank you for your contribution! I resolved the conflicts coming from the large UTF8 PR and did some minor cosmetic changes to the identifiers. The change is now good to go from my side.

Best,
Daniel

PR #620, opened 2026-05-26
feat: Emit declaration specific output

daogrady:
# Emit Declaration-Specific Output for `.d.ts` Files

This avoid breaking the .js files generated by cds-typer when users would transpile the accompanying .ts files.

### New Feature / Refactor

✨ Introduces proper `.d.ts` declaration file output by adding format-aware TypeScript printing via two distinct printer classes, `DeclarationPrinter` and `ImplementationPrinter`. The default for `output_d_ts_files` is changed from `false` to `true`, meaning generated files will now be proper type declarations by default.

The core change splits the monolithic printing logic in `visitor.js` into composable printer abstractions that conditionally emit correct syntax for either `.ts` (implementation) or `.d.ts` (declaration) output — omitting runtime-only constructs like `Object.defineProperty`, `static` + `declare` modifiers, and value initializers where they are not valid in declaration files.

### Changes

* `lib/printers/typescript.js` (**new file**): Introduces a `TypeScriptPrinter` base class with two concrete implementations — `ImplementationPrinter` (for `.ts`) and `DeclarationPrinter` (for `.d.ts`) — handling aspect functions, enum consts, `kind` properties, singular/plural class printing, events, and services.
* `lib/visitor.js`: Refactored to delegate output generation to `this.tsPrinter` (chosen based on `outputDTsFiles`). Removes inline printing methods (`#printStaticActions`, `#printStaticKeys`, etc.) and replaces them with calls to the printer. Guards `Object.defineProperty` calls behind `!configuration.outputDTsFiles`. Moves `MAX_TRANSITIVE_RESOLUTION_STEPS` and transitive type resolution to `resolver.js`.
* `lib/printers/wrappers.js`: Renames all exported wrapper functions (e.g., `createArrayOf` → `arrayOf`, `createUnionOf` → `unionOf`) to cleaner names without the `create` prefix. All consumers updated accordingly.
* `lib/file.js`: `SourceFile` and `FileRepository` now accept and propagate a `tsPrinter` instance. `printEnum` calls now require the printer to be passed. Improved error handling in `writeout` using `LOG` instead of `console.error`.
* `lib/components/enum.js`: `printEnum` now requires a `printer` parameter to delegate enum const printing.
* `lib/components/property.js`: `getPropertyModifiers` now returns `[]` (no `declare`) when `outputDTsFiles` is `true`.
* `lib/resolution/resolver.js`: Extracted `resolveTransitive` method for multi-level type reference resolution (previously inlined in `visitor.js`). Updated wrapper import names.
* `lib/typedefs.d.ts`: Added `traverser` module types (`EventPayload`, `Handler`, `Selector`, `HandlerOptions`) and new `printer` module types (`PrintContext`, `EnumPrintContext`).
* `lib/traverser.js`: Refactored typedef references to use the new `typedefs.d.ts` traverser types; `#emit` now supports an `options` parameter.
* `package.json`: Changed default value of `output_d_ts_files` from `false` to `true`.
* `CHANGELOG.md`: Documents the behavior change and new default for `output_d_ts_files`.
* `test/ast.js`: Extended AST visitor to handle `.d.ts`-style function declarations (no body, TypeLiteral return type) and synthesize initializers for const declarations in declaration files. Updated `getAspectFunctions`, `getAspects`, and `getInlineClassDeclarations` to support both formats.
* `test/unit/*.test.js`: Various tests updated to use `astw.getAspect()` helper, adjust static/readonly member checks for the new `.d.ts` default, or explicitly opt into `outputDTsFiles: false` where `.ts`-style AST structure is expected.

- [ ] 🔄 Regenerate and Update Summary

<details>
<summary>PR Bot Information</summary>

**Version:** `1.22.9`

- LLM: `anthropic--claude-4.6-sonnet`
- File Content Strategy: Full file content
- Event Trigger: `pull_request.edited`
- GithubContextProvider: [https://github.com/cap-js/cds-typer/pull/621](https://github.com/cap-js/cds-typer/pull/621)
- Correlation ID: `6dd4829c-6f21-47f0-bce0-20c16c95e852`
- Output Template: [Default Template](https://github.tools.sap/intelligent-insights/i2-pull-request/blob/main/src/services/llm/prompts/summary_default_output_template.md)
- Summary Prompt: [Default Prompt](https://github.tools.sap/intelligent-insights/i2-pull-request/blob/main/src/services/llm/prompts/summary_instructions_prompt.md)
</details>


daogrady:
Note: build script can stay as is for now. Part of it becomes a no-op, while other parts are still relevant. Especially for users setting `output_d_ts_files=false`, this ensures the build still runs properly.

chgeo (commented):
Great work 👍

Please add a changelog with a short info on what the benefit is for the users.  This doesn't come across in the PR description (so maybe add a paragraph there, too).

daogrady:
Fair point! I extended on it in the changelog and added a short summary in the PR description.

chgeo (approved):

PR #653, opened 2026-08-24
fix: Inline Enums in JS Files

daogrady:
Fixes cap/issues/issues/19670

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -320,6 +320,9 @@ class SourceFile extends File {
      *  (so `module.exports.Entity.prop = { ... }` is still emitted) but suppresses the TypeScript
      *  `const`/`type` declaration. Use this for projected entities whose enum is already declared
      *  in the source namespace file and should be referenced from there rather than re-declared.
+     * @param {boolean} [options.skipJsRegistration] - when true, suppresses the JS `Entity.prop ??= {...}`
+     *  assignment entirely. Use this for pure CDS aspects, which have no runtime JS value and therefore
+     *  cannot be used as an LHS in the assignment.
      * @example
      * ```js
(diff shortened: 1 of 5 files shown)


lib/file.js
```diff
@@ -320,6 +320,9 @@ class SourceFile extends File {
      *  (so `module.exports.Entity.prop = { ... }` is still emitted) but suppresses the TypeScript
      *  `const`/`type` declaration. Use this for projected entities whose enum is already declared
      *  in the source namespace file and should be referenced from there rather than re-declared.
+     * @param {boolean} [options.skipJsRegistration] - when true, suppresses the JS `Entity.prop ??= {...}`
```
chgeo, line 323: Why do we need the switch at all?

daogrady: To handle aspects that contain inline enums:

```cds
// schema.cds
aspect Status {
    status : String enum { Started; }
}
```
inline enums are made available during runtime with the following pattern:

```js
// index.js
module.exports.Status = ... // entity definition via cds.entities, wrapped in a proxy
module.exports.Status.status_enum =  ... // attach enum values
```
but _aspects_ get a TypeScript representation (index.ts), but none in JavaScript land (index.js), so it would lead to:

```js
// index.js
module.exports.Status.status_enum =  ... // attach enum values to undefined
//                   ^ ❌ NPE
``` 
once the aspect materialises in an actual entity, the enum is attached to the entity:

```cds
// schema.cds
extend MyEntity with Status
```

```js
// index.js
module.exports.MyEntity = ... // entity definition via cds.entities, wrapped in a proxy
module.exports.MyEntity.status_enum =  ... // enum values
```

chgeo (approved):

